PR #21, opened 2020-09-28
Testnet

djpnewton:
Using the changes in https://github.com/sparrowwallet/drongo/pull/3 we add the option for the user to choose bitcoin or testnet network.

todo:
 - reduce the amount of dialogs the user needs to go through (could probably merge the network diaglog with file open dialog etc..)
 - remove network dialog when we can infer which network it is (the most glaring one is when the user double clicks to view a wallet transaction in detail)
 - test with hardware wallets and HWI
 - figure out how to match this functionality with the electrum server settings
   - multiple electrum connections?
   - only allow a single network type for wallets during one session?


craigraw:
Thanks for the comprehensive PR! I actually started working on testnet support yesterday, so this has come at a good time.

A key question in the implementation is whether Sparrow should support mainnet usage alongside testnet usage "in the same session". While there are no doubt some use cases for this, it greatly increases the complexity of the implementation and substantially the user interface as well. 

For example, your implementation has the Network object passed around in order to understand what this state is whenever an address needs to be created. bitcoinj does this as well. However, if we assume the state is constant for a runtime session then a static variable (which is protected from any but the initial assignment) would be much simpler. 

Similarly, supporting multiple Electrum server connection setups seems be substantially increasing the complexity of the server preferences and other dialogs (wallet setup, import etc). If however we refuse to open/import wallets with xpubs in testnet and tpubs in mainnet (with an appropriate error message) much of the complexity is removed. I also have some concern about the increased potential for user confusion when mixing testnet and mainnet wallets.

Running testnet in a separate bitcoind/electrumx/electrs process is of course required, so there is precedent for this mode of operation. That's not to say Sparrow must follow, but there would need to be a significant benefit for doing so. I'm not seeing this benefit outweighing the cost right now. What are your thoughts?

djpnewton:
I think you are probably right that having the sparrow process run in bitcoin or testnet mode and not accept/create wallets in the wrong mode is the better idea.

Using a command line or environment flag to trigger testnet mode and then reading a different config file path (like bitcoind) could work well and would enable the user to easily use different electrum servers when changing from bitcoin to testnet.

Do you think the static Network setting should be in the Drongo library or just in the Sparrow app (which I guess would allow the drongo library to be more flexible in how its used)?

craigraw:
Thanks, good to get that feedback. I'll implement a command line parameter to set a different .sparrow folder, and put the network setting in the config file. This should make it simple to switch back and forth on the command line with appropriate wallet subfolders for both networks etc.

Re the static setting, I'm putting it in the Drongo library, as not doing so means carrying the network state around in local variables in many different places. It would certainly be more flexible to do the latter, but again implementation complexity/maintenance is the key concern given Drongo is likely to focus on supporting Sparrow. I'd like both Drongo and Sparrow to have the same mental model when it comes to networks.

djpnewton:
Ok sounds good.

Will you put a network setting in the wallet file? Or just rely on them being in different sub-directories?

craigraw:
I think it will need to be added to the wallet file. Although Sparrow already gives an error when it encounters a tpub in a mainnet config and vice-versa, there is no such check for a BIP39 seed keystore. The setting will default to the currently configured network when not present in the wallet file.

It also makes sense to follow the bitcoind (and many other) precedent of creating a testnet folder in the .sparrow folder for testnet config and wallets when a sparrow home folder is not explicitly configured on the command line.

PR #90, opened 2021-04-05
Update actions

w0wn3r0:


PoloLacoste:
Be carefull you have to block the user !
It's still mining right now...

craigraw:
Thanks. Have blocked and reported this user.

Have also changed access to local Github Actions only. Should be the default!

PR #132, opened 2021-05-25
Close "About" and "Introduction" when Escape key is pressed

haakonn:
This closes the Introduction and About windows when the user presses Escape.

I put a utility method for this into `AppServices` for lack of a better place; please let me know if it should go somewhere else.


craigraw:
Looks good, thanks!

PR #133, opened 2021-05-26
Fixes around transaction saving menu items

haakonn:
Three related minor issues are addressed:

1. `File -> Save transaction` and `File -> Save PSBT -> As Binary` share the Ctrl+S accelerator, but only one of the items should be enabled at a time. It turns out that even if the parent `File -> Save PSBT` menu is disabled, the `As Binary` accelerator is still active and gets priority over `Save transaction`. This PR explicitly disables `As Binary` along with `File -> Save PSBT` to address this.

2. When Sparrow starts up without any tabs open, the `Save transaction` menu item is enabled, but clicking it makes no sense and causes a `NullPointerException`. This PR disables it on startup. As far as I can tell, there is no session management or similar that would be disrupted by this.

3. When the last tab to be closed is a transaction tab (such that there are no more tabs open), the `File -> Save PSBT` menu item is still visible and enabled, and will cause a `NullPointerException` if selected. This PR fixes that.

### DIFF
--- a/src/main/java/com/sparrowwallet/sparrow/AppController.java
+++ b/src/main/java/com/sparrowwallet/sparrow/AppController.java
@@ -88,6 +88,9 @@ public class AppController implements Initializable {
     private Menu savePSBT;
 
     @FXML
+    private MenuItem savePSBTBinary;
+
+    @FXML
     private MenuItem exportWallet;
 
     @FXML
@@ -267,6 +270,7 @@ public class AppController implements Initializable {
         showLoadingLog.setSelected(Config.get().isShowLoadingLog());
         showUtxosChart.setSelected(Config.get().isShowUtxosChart());
         savePSBT.visibleProperty().bind(saveTransaction.visibleProperty().not());
+        savePSBTBinary.disableProperty().bind(saveTransaction.visibleProperty());
         exportWallet.setDisable(true);
         refreshWallet.disableProperty().bind(Bindings.or(exportWallet.disableProperty(), Bindings.or(serverToggle.disableProperty(), AppServices.onlineProperty().not())));
         sendToMany.disableProperty().bind(exportWallet.disableProperty());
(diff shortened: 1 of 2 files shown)


src/main/java/com/sparrowwallet/sparrow/AppController.java
```diff
@@ -232,6 +235,7 @@ void initializeView() {
                 if(tabs.getTabs().isEmpty()) {
                     Stage tabStage = (Stage)tabs.getScene().getWindow();
                     tabStage.setTitle("Sparrow");
+                    saveTransaction.setVisible(true);
```
craigraw, line 238: I think there needs to be an additional line below this:
```
saveTransaction.setDisabled(true);
```

craigraw:
Thanks for this.

Oddly enough I can't reproduce the first issue on OSX - assume you are testing on Linux or Windows? At any rate the changes for it seem good to have anyway.

src/main/java/com/sparrowwallet/sparrow/AppController.java
haakonn: Thanks, wasn't sure. Added it now.

haakonn:
I am indeed on Linux, and haven't tested on other systems.

I forget exactly how to reproduce the first issue, but one way to see it manifest on current master is to start Sparrow without any open tabs. Now `File -> Save Transaction` is enabled and has the Ctrl+S accelerator, but if you press Ctrl+S, you get an NPE from `AppController.savePSBTBinary`, which is the `Save PSBT -> As Binary` action. At least I do.

craigraw:
Interestingly I confirmed OSX and Linux do differ using this test. OSX NPEs on `saveTransaction` while Linux NPEs on `savePSBT`.

Merging this now.

haakonn:
Thanks for taking the time to confirm that! Strange, I wouldn't have expected platform differences there.

PR #134, opened 2021-05-27
Make mnemonic entry more efficient

haakonn:
I was bothered by a couple of issues when entering mnemonics:

1. The autocompletion dropdown needs an extra keypress to close. After selecting a word from the dropdown, you have to press TAB twice (or ENTER and then TAB) in order to move to the next word field.
2. I was confused when entering words for which there are other words that are full prefixes. For example, when you want to enter "label", and you have typed the first three letters, the completion drowdown closes because "lab" is also in the wordlist. Same with "lawn vs "law" etc. Then when you enter the fourth letter, the dropdown reappears.

This PR keeps the dropdown open until a word is chosen, and also immediately moves to the next field upon completion. I find this a much more efficient way to type in mnemonics, and more in line with other wallet software.


craigraw:
Good improvement!

PR #179, opened 2021-08-22
Fix when getProxyServer returns null on getWhirlpool

diogorac:
I wanted to try out the whirlpool, but I was getting an empty list of available pools.

State:
* Testnet
* No proxy server set
* Mixing UTXOs from a wallet generated by a previous [non whirpool] sparrow version.

I took a look, and `getProxyServer()` was returning `null`, and throwing an error. So I completed the condition, alongside with checking if it's empty. (I don't know if it can happen).


craigraw:
Thanks, good catch. The proxy server can indeed be `null` with a new config. 

I've made the change separately for simplicity - it will be committed today.

craigraw:
Change has now been committed, along with a lot of additional work. Any further testing appreciated!

Issue #188, opened 2021-09-06
Reproducible build of 1.4.3 did not match 


achow101:
I attempted to perform a reproducible build of Sparrow 1.4.3 but my build results did not match the published released. The build was deterministic in that repeated attempts to build it resulted in the same binary, but the build results did not match the published binaries.

It seems like the problem is that the Java runtime is bundled with Sparrow, and this runtime is pulled from the system doing the build itself. So I likely have a different runtime installed and this causes the non-reproducibility.

Additionally, the deb and rpm files are not reproducible. Diffoscope tells me that this is partially timestamp based, but also more complicated than just that.

craigraw:
Thanks for the investigation into this. Reproducibility is very much a goal for this project.

As you note, v1.4.3 is not reproducible across different machines. Versions earlier than this were in fact not even reproducible on the same machine, but the upgrade to Java 16 (which was largely done to resolve this reproducibility related bug: https://bugs.java.com/bugdatabase/view_bug.do?bug_id=JDK-8252730) resolved this.

The runtime bundled with Sparrow can be problem in achieving reproducibility, but this can be resolved by specifying the same Java release by the same vendor. For Sparrow, this is currently AdoptOpenJdk jdk-16.0.1+9 Hotspot (available for all supported platforms from this page: https://adoptopenjdk.net/archive.html?variant=openjdk16&jvmVariant=hotspot). For future reference, they are:
- [Linux x64](https://github.com/AdoptOpenJDK/openjdk16-binaries/releases/download/jdk-16.0.1%2B9/OpenJDK16U-jdk_x64_linux_hotspot_16.0.1_9.tar.gz)
- [MacOS x64](https://github.com/AdoptOpenJDK/openjdk16-binaries/releases/download/jdk-16.0.1%2B9/OpenJDK16U-jdk_x64_mac_hotspot_16.0.1_9.tar.gz)
- [Windows x64](https://github.com/AdoptOpenJDK/openjdk16-binaries/releases/download/jdk-16.0.1%2B9/OpenJDK16U-jdk_x64_windows_hotspot_16.0.1_9.zip)

When `./gradlew jpackage` is executed running on this release (use `./gradlew -v` to check) then the included Java runtime will be identical.

[...]

PR #191, opened 2021-09-07
Update required Java version in README

romanz:
Building with Java 14 results in the following error:

```
./sparrow/src/main/java/com/sparrowwallet/sparrow/whirlpool/dataSource/SparrowWalletStateSupplier.java:64: error: pattern matching in instanceof is a preview feature and is disabled by default.
            if(externalDestination.getPostmixHandler() instanceof SparrowPostmixHandler sparrowPostmixHandler) {
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,7 +28,7 @@ If you prefer to run Sparrow directly from source, it can be launched with
 
 `./sparrow`
 
-Java 14 or higher must be installed.
+Java 16 or higher must be installed.
 
 ## Configuration
 


romanz:
I've updated to Java 16 following this comment: https://github.com/sparrowwallet/sparrow/issues/188#issuecomment-914044370

craigraw:
Thanks :) Fixed in 44b9cee. I also added a link to the JDK version used to build the release binaries.

PR #195, opened 2021-09-09
Fix skdman link

RequestPrivacy:
The current link takes me to `https://github.com/sparrowwallet/sparrow/blob/master/docs/sdkman.io` with an "404 not found error" message.


craigraw:
Thanks!

PR #235, opened 2021-10-04
Allow running `sparrow` script from other paths

romanz:
IIUC, it currently assumes that it's run locally using `./sparrow`.


craigraw:
Thanks!

Issue #263, opened 2021-10-14
Struggling to get the command line working


sashafklein:
This seems like a dumb request, but I'm unclear on something. 

I've installed Sparrow "normally", to both a Mac and Linux computer. It works just fine. 

The docs say that to run in testnet (which I would like to do), I need to run it with flags on the command line, but on both computers, the command line command "sparrow" or "sparrow" throws a `command not found` error. Am I misunderstanding this? Do I need to clone the repo to run that command?

Generally, think it would be helpful to add instructions on _how_ to run Sparrow from the command line to the docs. (I'd be happy to write this up, if someone could point me in the right direction).

Thanks for the excellent wallet!

PR #267, opened 2021-10-15
Clarify how to run sparrow from CL

sashafklein:
This tiny Readme edit would, I think, have preempted the confusion that prompted #263 . I thought that the general install installed a global command, and didn't think, for some time, to run from within the directory. Dumb, but figured clarifying the doc would be better than just figuring it out silently on my own.

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,7 +31,7 @@ The release binaries are reproducible from v1.5.0 onwards (pre codesigning and i
 
 ## Running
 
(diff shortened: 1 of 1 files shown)


RequestPrivacy:
Maybe going forward it would be better to automatically add the `sparrow` file directory to the users `PATH` during the first run so he can start Sparrow from anywhere? Rather then forcing him to go to the project directory/create an alias/add the folder manually to his `PATH`?

PR #302, opened 2021-11-09
Fix typo in error message

jooray:


craigraw:
Thanks, fixed as part of fb85277.

PR #308, opened 2021-11-13
grammar and spelling updates in the about dialog

aphex3k:


craigraw:
Thanks! I didn't agree with all the corrections, but I've taken several of them into d1d1b08.

Issue #321, opened 2021-11-22
[Feature] Add option to set custom fee, i.e. flat fee of 1 sat/vbyte


dergigi:
I'd love to have an additional option that says "custom" where I just can the fee to whatever I want. 

<img width="394" alt="Screenshot 2021-11-22 at 21 40 49" src="https://user-images.githubusercontent.com/109058/142932189-650deff5-d57e-4f44-85ab-a39426616719.png">

**Reasoning / scenario:**
* User wants to broadcast a lot of transactions and knows (or hopes) that the mempool is empty / will clear
* Currently, the user has to set the slider to 1sat for every transaction
* With this feature the user can set it once and the fee slider will default to 1sat

craigraw:
Added in 1b61a78. A fee rate source of "Minimum (1 sat/vB)" is now present in the drop-down for Fee rates in the General Preferences. If the Send tab is already open, press Clear to revert to the current fee rate source.

PR #324, opened 2021-11-23
Add option to set flat fee of 1 sat/vbyte

alaznem:
As suggested by @dergigi in issue #321.

This pr contains the cheapest solution I could think of.

**Privacy concern**: It leaks application usage data to github, because it queries a static file from github.

The new static fee rate source option queries a raw json file from /src/main/resources/com/sparrowwallet/sparrow/preferences/feeRatesSource-static-1-sat-per-vbyte.json

If this solution doesn't meet the code quality standard of sparrow, maybe someone can use it as starter and solve the data source issue for the json more elegant.


craigraw:
Thanks for this - added in 1b61a78 without the need for the external call.

PR #363, opened 2021-12-16
logback vuln

ladams1776:
Hi there, can close this PR and make an official one if mine is way off basis. But I saw this:

I see logback has a vuln, not exact, but similar to log4j: http://slf4j.org/log4shell.html 

"Logback does NOT offer a lookup mechanism at the message level. Thus, it is deemed safe with respect to CVE-2021-44228.

However, logback may make JNDI calls from within its configuration file. This was recently reported in LOGBACK-1591 as a vulnerability of lesser severity. In response, we have released logback version 1.2.8. Please upgrade."

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -360,7 +360,7 @@ extraJavaModuleInfo {
     module('json-simple-1.1.1.jar', 'json.simple', '1.1.1') {
         exports('org.json.simple')
     }
-    module('logback-classic-1.2.3.jar', 'logback.classic', '1.2.3') {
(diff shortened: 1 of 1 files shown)


craigraw:
Thanks for this suggestion - although the risk is low it's wise to make this update.

Because the logback dependency is declared in the submodule drongo, I've gone ahead and changed it there and in Sparrow (see https://github.com/sparrowwallet/drongo/commit/645be6a5a5a7b434ea9a3991ae85e0b8b6d4f50b and 5ee9796).

PR #483, opened 2022-03-24
[feature request] pay-to-many with comma-separated string entry for large numbers of recipients

mchapiro1:
From what has been relayed to me, currently the process of constructing a pay-to-many transaction involves a clicking operation, which limits the number of recipients that are practical to add.

It would be nice to be able to paste in a string of all recipient data in one step (recipient1, amount1, receipient2, amount2,...), which I think Electrum allows.

In particular, I would like this in Sparrow so it can be implemented in conjunction with PayNyms.

This has potential privacy benefits since storing payee paynyms poses less of a risk in the event of a data breach even if in theory without such a breach, the payee privacy stays constant.

[edit: I don't know what the bwt branch is, but it was the only way I was able to do a PR. I'm not familiar with GitHub so hopefully this can be moved into the appropriate place]


craigraw:
I'm unable to merge this, or even see what is changed - there are far too many conflicts as the bwt branch you are working off is very old. 

Re pasting a string, why not just create and load a csv?

Re paying PayNyms, it's a possibility, although not straightforward if they are not already linked. You can add an issue for this though to track it.

PR #567, opened 2022-05-23
Rename the passphrase label to salt

1ma:
A minor UX improvement suggestion. I've been baffled while trying to import a BIP-39 wallet, thinking that "passphrase" was the password for the wallet file encryption at rest, so I was getting an unexpected set of addresses.


craigraw:
Thanks for the suggestion, but [BIP39](https://github.com/bitcoin/bips/blob/master/bip-0039.mediawiki) defines this field as a passphrase, and that's what it's generally called. Renaming it 'salt' would, despite the advantage of a different looking name to 'password', cause more confusion than it relieves.

1ma:
Fair enough, I wasn't aware these names were specced.

PR #675, opened 2022-08-29
improve reproducible docs

nyxnor:
- apt-key add is deprecated, use /usr/share/keyrings and [signed-by=/path/key]
- download key securely and verify digital signature
- correct headers
- default clone to https://, not git@
- explain each step of adding the deb repo for the user to understand what is being done, so no mystery for new users

### DIFF
--- a/docs/reproducible.md
+++ b/docs/reproducible.md
@@ -1,18 +1,21 @@
-## Reproducible builds
+# Reproducible builds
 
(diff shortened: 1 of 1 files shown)


craigraw:
Good improvements, thanks for this.

I tend to use SDKman myself, so I am less familiar with the Debian repo approach.

PR #728, opened 2022-10-30
Add Sparrow Server Dockerfile

sethforprivacy:
This Dockerfile downloads, verifies, and allows the usage of Sparrow Server via Docker.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -37,7 +37,7 @@ RUN case ${TARGETARCH:-amd64} in \
     | grep -q "GOODSIG ${PGP_SIG}" \
     || exit 1 \
     && sha256sum --check sparrow-1.7.0-manifest.txt --ignore-missing || exit 1 \
-    && tar xvf sparrow-server-${SPARROW_VERSION}-${SPARROW_ARCH}.tar.gz -C /opt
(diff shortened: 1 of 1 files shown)


craigraw:
Thanks for this @sethforprivacy. Do you think it's worthwhile having this here as well as in https://github.com/sethforprivacy/sparrow-server-docker? I'm always mindful of keeping the scope of maintenance on the project to a manageable level.

sethforprivacy:
I prefer to have it upstream as well as an example and in case you want to have docs for building Sparrow via Docker, but it's really your final decision.

I'll work to keep my image up to date, so if you don't want it in the repo, feel free to reject the PR!

benpbolton:
I could see a link to https://github.com/sethforprivacy/sparrow-server-docker in the README or docs for those interested in a verified Docker implementation, but agree that keeping scope focused is probably the bigger win.

suurt8ll:
@sethforprivacy seems to have archived his repository. Just giving heads up for future readers.

sethforprivacy:
> @sethforprivacy seems to have archived his repository. Just giving heads up for future readers.

Note that the Dockerfile is still useful, but I don't see any point in maintaining it as the only reason I know of for people to use sparrow-server was to be able to remix easily. If there is a clear need I'm happy to revive it.

Issue #619, opened 2022-07-07
Unable to reproduce runtime folder of 1.6.5 build


charliecharles1234:
I am trying to reproduce the release binary for 1.6.5 as per [the instructions](https://github.com/sparrowwallet/sparrow/blob/master/docs/reproducible.md), but my `lib/runtime/lib` folder differs.

I have recorded my attempt with Asciinema: https://asciinema.org/a/ixn6XcKyJ9rVa7VOHKzcrxsNL

This is run inside a new Ubuntu Jammy amd64 container, with the only activities before the beginning of the recording being installing a few necessary packages e.g. asciinema.

Right at the end I verified the java version to check that it is correct.

craigraw:
Thanks for the recording. The majority of the differing files relate to the JVM. As far as I can see, the installation process for Java 17.0.2 is correct (and matches my own approach, and my own `java --version` output). I've just run the reproduction build on Ubuntu 20.04 and all files matched, except for `lib/runtime/lib/modules`, which is a known issue relating to a bug in Java (see #197).

Wrt the JVM files, I'm perplexed as to why this is occurring. The only difference I can see is the difference in Ubuntu versions (which should make no difference at all). Is there anything about this container that might cause sdkman to download a different architecture version of Java 17.0.2?

```
uname -a
Linux xxx 5.13.0-48-generic #54-20.04.1-Ubuntu SMP Thu Jun 2 23:37:17 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux
```

RequestPrivacy:
Can reproduce all the diffs on v1.6.6

```
requestprivacy@kubuntu:~/Bitcoin$ java -version
openjdk version "18.0.1" 2022-04-19
OpenJDK Runtime Environment Temurin-18.0.1+10 (build 18.0.1+10)
OpenJDK 64-Bit Server VM Temurin-18.0.1+10 (build 18.0.1+10, mixed mode, sharing)
```

```
requestprivacy@kubuntu:~$ uname -a
Linux kubuntu 5.15.0-43-generic #46-Ubuntu SMP Tue Jul 12 10:30:17 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux
```

```
requestprivacy@kubuntu:~/Bitcoin$ diff -r ./sparrow/build/jpackage/Sparrow ./Sparrow
[...]

PR #730, opened 2022-11-01
Add git pull & git submodule update --checkout

RequestPrivacy:
Adds an additional guideline for people having the repo already cloned.

Further, the PR tries to mitigate the risk of failed reproducible builds due to wrong commit checked out at the drongo submodule, see problem description https://github.com/sparrowwallet/sparrow/issues/619#issuecomment-1228697555 or [https://stackoverflow.com/questions/1899792/why-is-git-submodule-not-updated-automatically-on-git-checkout](https://stackoverflow.com/questions/1899792/why-is-git-submodule-not-updated-automatically-on-git-checkout)

For reference of  `git submodule update --checkout` see [https://stackoverflow.com/a/41920893](https://stackoverflow.com/a/41920893)

For a permanent solution on the users side [https://stackoverflow.com/questions/4611512/is-there-a-way-to-make-git-pull-automatically-update-submodules](https://stackoverflow.com/questions/4611512/is-there-a-way-to-make-git-pull-automatically-update-submodules) provides additional hints.

### DIFF
--- a/docs/reproducible.md
+++ b/docs/reproducible.md
@@ -94,6 +94,7 @@ git clone --recursive --branch "${GIT_TAG}" https://github.com/sparrowwallet/spa
 If you already have the sparrow repo cloned, fetch all new updates and checkout the release. For this, change into your local sparrow folder and execute:
 
 ```shell
+cd {yourPathToSparrow}/sparrow
 git pull --recurse-submodules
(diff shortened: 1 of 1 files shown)


craigraw:
Thanks - this is a good improvement!

PR #736, opened 2022-11-05
Add Bylls BTC/CAD rates

misnor:
Adds CAD support for Bylls exchange rates.

### DIFF
--- a/src/main/java/com/sparrowwallet/sparrow/net/ExchangeSource.java
+++ b/src/main/java/com/sparrowwallet/sparrow/net/ExchangeSource.java
@@ -107,6 +107,48 @@ public enum ExchangeSource {
                 return new CoinGeckoRates();
             }
         }
+    },
+    BYLLS("Bylls")
+    {
+        @Override
+        public List<Currency> getSupportedCurrencies() {
+            // Bylls is a CAD only exchange
+            String currency = getRates().public_price.to_currency;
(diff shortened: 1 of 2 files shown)


craigraw:
Thanks for the PR.

I may be missing something, but why add an exchange source with single currency? Seems to be a poor functionality to maintenance tradeoff. Is there something that the exchange sources do not provide wrt CAD?

misnor:
Hey - thanks for the quick response!

Bylls/Bull is a non-custodial Canadian exchange. [Their FAQ is pretty extensive](https://bullbitcoin.com/faq#about-faq-557104669) if you have a moment.

The CEO (Francis) previously added Bylls BTC to [Electrum in 2019](https://github.com/spesmilo/electrum/pull/5211/files) - I am using the same route he used then, so I believe the API route is stable.

A use case that isn't great: If a user is going through all the exchanges on any currency other than CAD, (say "USD" or "JPY") when they select Bylls the currency will switch to CAD - now switching back to coinbase the currency will still be CAD.

This PR is very much just me sharing a patch of some functionality I put together this weekend, I won't be upset if it doesn't fit within the repository!

craigraw:
Thanks, appreciate the share. I'll certainly keep Bylls in mind should additional exchange sources be required in future, but I think it's probably not a great fit for Sparrow today.

PR #743, opened 2022-11-08
Add splash-screen during start-up

RequestPrivacy:
#### Description
It would be nice to have a short splash-screen before confronting the user with the GUI. As Sparrow hasn't much work to do before showing the GUI, this is only for user experience at this point. 

#### Work outstanding
- [ ] FXML design of the splash screen itself
- [ ] Decide if splash screen should be implemented in the Preloader like in the first commit or rather in the start() method of the GUI application - mainly due to both, Terminal and GUI, using the same Preloader right now.

Is a splash-screen something you might want to consider for Sparrow?

### DIFF
--- a/src/main/java/com/sparrowwallet/sparrow/SparrowWalletPreloader.java
+++ b/src/main/java/com/sparrowwallet/sparrow/SparrowWalletPreloader.java
@@ -1,11 +1,46 @@
 package com.sparrowwallet.sparrow;
 
 import javafx.application.Preloader;
+import javafx.fxml.FXMLLoader;
+import javafx.scene.Scene;
 import javafx.stage.Stage;
+import javafx.stage.StageStyle;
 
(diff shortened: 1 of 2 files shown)


craigraw:
> Is a splash-screen something you might want to consider for Sparrow?

I'm not sure. Generally a splash screen is there to tell the user something is going on while the application loads. Hopefully, Sparrow doesn't require that kind of feedback because it loads quickly enough.

RequestPrivacy:
I guess from a pure technical/functional standpoint you're right. 

From the user experience side...I don't know. On the one hand there is the "snappy" feeling if Sparrow simply pops up, on the other hand - I at least - find it comforting to see something like a "welcome screen".

But I think we both value minimal code for maximal security/maintainability, so without more people demanding something like that the benefits don't out-weight the technical burden. Closing this off.

Issue #598, opened 2022-06-13
[UI] Whirlpool loading bars indentation with different mix counts


RequestPrivacy:
# Description 

The loading bars could be indented more consistently if there are UTXOs with different mixing counts (e.g. with utxos with one digit and two digit mixing history like 9 and 10 the loading bar seems to be intended by one digit?!), also see the "number 1"-labeling in the screenshot provided in #597.

```
Mixes
 --------loading bar--------    9
--------loading bar--------    10
```

Maybe it is possible to have them exactly below each other even if the digit count of the numbers are different or else maybe change to a numbering system with a fixed amount of digits like maybe 4:

```
Mixes
--------loading bar--------    0004
--------loading bar--------    0200
--------loading bar--------    0010
--------loading bar--------    4000
```

PR #748, opened 2022-11-14
Add uniform string length formatting to MixStatusCell text

RequestPrivacy:
Intended to fix issue #598 by setting the string representation of each MixStatusCell to the length of the the length of the highest `MixStatus.getMixesDone()` encountered. This way the fixed-width gap to the ProgressBar graphic to the left results in the ProgressBars being aligned with each other in the "Mixes" column of Pre-& Postmix wallets UTXO tab.

I'm still figuring out my test environment, so would be glad for feedback/help on:

- [ ] Do the white-spaces added represent the width of a digit (i.e. are all ProgressBars aligned for all denominations and `mixesDone` orders of magnitude)?
- [ ] Are the ProgressBars aligned for Premix- & Postmix wallets?
- [ ] Are all ProgressBars updated immediately if the highest order of a `mixesDone` changes e.g. from 9 -> 10?

### DIFF
--- a/src/main/java/com/sparrowwallet/sparrow/control/MixStatusCell.java
+++ b/src/main/java/com/sparrowwallet/sparrow/control/MixStatusCell.java
@@ -21,6 +21,7 @@ import java.util.Locale;
 
 public class MixStatusCell extends TreeTableCell<Entry, UtxoEntry.MixStatus> {
     private static final int ERROR_DISPLAY_MILLIS = 5 * 60 * 1000;
+    private static int maxMixesDoneDigitCount = 0;
 
     public MixStatusCell() {
         super();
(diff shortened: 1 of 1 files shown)


craigraw:
Thanks for the PR.

The font used to display the mix count is unfortunately not a fixed width font, so this approach won't work (the digits are not consistent widths). There is a potential solution, but it requires a more complex rewrite of the mix status cell.

RequestPrivacy:
Yeah I guessed this could pose a problem during my rudimentary testing. Out of curiosity: do you intend to split the cell further into two parts, one for the graphic and one for the text?

kmk142789 (approved):

PR #752, opened 2022-11-18
Updated Passport Single/Multisig Import and Export Instructions

BitcoinQnA:
Suited to the updated Batch 2 UI.


craigraw:
Thanks!

PR #753, opened 2022-11-18
Update bitcoin.lukechilds.co to bitcoin.lu.ke

lukechilds:
I migrated my services over to a new domain, I'll still be keeping the old domain running so old versions of software using it keep working but it would be great to get my new domain in the latest versions.

Should I also update `LUKECHILDS_CO` to `LU_KE` or will that forget the setting for existing users?


craigraw:
Thanks!

> Should I also update LUKECHILDS_CO to LU_KE or will that forget the setting for existing users?

It will forget the setting, which is minor in this case, but there's no real reason to change it - the enum name is not used in the display in any way.

PR #803, opened 2023-01-06
Added note to README indicating that Java 19 isn't yet supported

wombat6:


craigraw:
This is actually due to Gradle - I've upgraded the Gradle wrapper to 7.6 in 8a51a471, and Sparrow now builds on Java 19.

PR #843, opened 2023-02-15
add StartupWMClass to linux .desktop file

secondl1ght:
As the title says I added this so that when you add Sparrow to your application dash bar and open a new window, it is now associated with the already present application icon and a temp icon is not generated in the dash.


craigraw:
Thanks, good addition.

PR #860, opened 2023-02-24
Add context menu to copy amount values from amount cells

wazint:


craigraw:
I've avoided doing this, because of the cost of adding a context menu to every amount cell - potentially many thousands. That said, the performance impact would need to be properly benchmarked.

Issue #996, opened 2023-06-21
message id in Electrum server request 


santochibtc:
Why is a path included in the message id in this request to Electrum server?
https://github.com/sparrowwallet/sparrow/blob/c4651025bedf3e3fd881980c2ffe42c6ce49a002/src/main/java/com/sparrowwallet/sparrow/net/SimpleElectrumServerRpc.java#LL127C25-L127C45
This is an example request sent by Sparrow
{"jsonrpc":"2.0","method":"blockchain.scripthash.subscribe","params":["....."],"id":"m/0/0-15"}
Is this correct?

craigraw:
Historical reasons. `SimpleElectrumServerRpc` is used rarely, as almost all Electrum servers support batching and thus use `BatchedElectrumServerRpc`. But there is no particular reason to add this to the request, so I've removed it in 2deab05c. Thanks for pointing it out.

PR #997, opened 2023-06-21
fix message id in electrum requests

santochibtc:
https://github.com/sparrowwallet/sparrow/issues/996#issue-1768138658


craigraw:
Thanks - I've used 2deab05c instead to preserve the indenting :)

PR #1147, opened 2023-11-02
Fix computation of master fingerprint for Satochip device

Toporin:


craigraw:
Thanks - tested and added in 910a400b.

PR #1187, opened 2023-11-30
Satochip: Patch error 0x9C01 in cardBip32GetExtendedKey()

Toporin:
This error occurs when the cache memory in the card is full. 
We need to flush the card cache by sending the same command with the p2 flag set to 0x80.


craigraw:
Thanks

PR #1204, opened 2023-12-22
feat: Implement bitcoin unit toggle on bitcoin balance

krzyczak:
Clicking the bitcoin balance toggles the units between sats and BTC.

sats-to-btc


craigraw:
Thanks for suggesting this. I think though that it may make sense to toggle all bitcoin unit fields, rather than just one?

craigraw:
Decided to leave it as is.

PR #1275, opened 2024-02-21
Sorted value column in Addresses

MisterSeven144:
Unless there's a good reason to prevent sorting on the `Value` column, I think it's useful to be able to sort the wallet's addresses by value. 🤞 

image

### DIFF
--- a/src/main/java/com/sparrowwallet/sparrow/control/AddressTreeTable.java
+++ b/src/main/java/com/sparrowwallet/sparrow/control/AddressTreeTable.java
@@ -60,7 +60,7 @@ public class AddressTreeTable extends CoinTreeTable {
             return new ReadOnlyObjectWrapper<>(param.getValue().getValue().getValue());
         });
         amountCol.setCellFactory(p -> new CoinCell());
-        amountCol.setSortable(false);
+        amountCol.setSortable(true);
         getColumns().add(amountCol);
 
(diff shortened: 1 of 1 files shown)


craigraw:
Addresses should be sorted by index IMO, and I can't see a good reason to do otherwise. Also, I haven't tested this fully, but I believe it has the potential to introduce subtle bugs in transaction loading. You'll also note that once you sort by value, you can't "undo" that sort - there is no way to return to the original "by index" sorting (I have no desire to introduce an index column).

MisterSeven144:
Thanks for the response.  I'm looking to view the addresses in my wallet by value so I can see the ones that have "a lot" of BTC, all the way down to those that barely have any.  If it doesn't make sense to do it on this screen (due to potential side-effects), is there another place in Sparrow to get this information?

craigraw:
Yes, the UTXOs tab.

MisterSeven144:
The problem is that I have several addresses in my wallet with multiple outputs.  So while the UTXO screen allows me to sort by the `Value` column, it's at the UTXO level.  What I'm hoping to see is a sorted list of addresses with their total value.

craigraw:
> The problem is that I have several addresses in my wallet with multiple outputs.

This is not ideal from a privacy perspective - to the extent that Sparrow is opinionated, it is opinionated in trying to dissuade from poor privacy practices.

Issue #1276, opened 2024-02-21
Exchange rates sometimes do not load


PrinceOfEgypt:
There is an intermittent issue where exchange rates do not load when you start sparrow and no fiat value is displayed:

![image](https://github.com/sparrowwallet/sparrow/assets/2649235/47fee711-e74e-4eb4-8f6a-978b02f9d1cd)

if you leave sparrow running for 5 minutes it will eventually load:

![image](https://github.com/sparrowwallet/sparrow/assets/2649235/29d8a6eb-9376-4e83-b8f9-866fb6376142)

when the issue occurs the following error is logged in sparrow.log (running in debug mode):

2024-02-21 14:20:44,700 WARN [Thread-7] c.s.s.n.ExchangeSource [null:-1] Error retrieving currency rates
com.samourai.wallet.api.backend.beans.HttpException: java.util.concurrent.ExecutionException: java.util.concurrent.RejectedExecutionException: HttpClient@1ed271cf{STOPPING} is stopped
	at com.sparrowwallet.nightjar@0.2.39/com.samourai.http.client.JacksonHttpClient.getJson(Unknown Source)
	at com.sparrowwallet.nightjar@0.2.39/com.samourai.http.client.JacksonHttpClient.getJson(Unknown Source)
	at com.sparrowwallet.sparrow@1.8.2/com.sparrowwallet.sparrow.net.HttpClientService.requestJson(Unknown Source)
	at com.sparrowwallet.sparrow@1.8.2/com.sparrowwallet.sparrow.net.ExchangeSource$3.getRates(Unknown Source)
	at com.sparrowwallet.sparrow@1.8.2/com.sparrowwallet.sparrow.net.ExchangeSource$3.getExchangeRate(Unknown Source)
	at com.sparrowwallet.sparrow@1.8.2/com.sparrowwallet.sparrow.net.ExchangeSource$RatesService$1.call(Unknown Source)
	at com.sparrowwallet.sparrow@1.8.2/com.sparrowwallet.sparrow.net.ExchangeSource$RatesService$1.call(Unknown Source)
	at javafx.graphics@18/javafx.concurrent.Task$TaskCallable.call(Unknown Source)
	at java.base/java.util.concurrent.FutureTask.run(Unknown Source)
	at javafx.graphics@18/javafx.concurrent.Service.lambda$executeTask$6(Unknown Source)
	at java.base/java.security.AccessController.doPrivileged(Unknown Source)
[...]

PR #1281, opened 2024-02-24
Add delay to RatesService startup (fixes #12760)

PrinceOfEgypt:
this is to address #1276 - I noticed that adding a startup delay to the ratesService caused to problem to go away, which leads me to believe that is some sort of concurrency issue. Please review as Java is not my primary language - there may be a better / more "correct" way to address the issue

### DIFF
--- a/src/main/java/com/sparrowwallet/sparrow/AppServices.java
+++ b/src/main/java/com/sparrowwallet/sparrow/AppServices.java
@@ -78,6 +78,7 @@ public class AppServices {
     private static final int RATES_PERIOD_SECS = 5 * 60;
     private static final int VERSION_CHECK_PERIOD_HOURS = 24;
     private static final int CONNECTION_DELAY_SECS = 2;
+    private static final int RATES_STARTUP_DELAY_SECS = 5;
     private static final ExchangeSource DEFAULT_EXCHANGE_SOURCE = ExchangeSource.COINGECKO;
     private static final Currency DEFAULT_FIAT_CURRENCY = Currency.getInstance("USD");
     private static final String TOR_DEFAULT_PROXY_CIRCUIT_ID = "default";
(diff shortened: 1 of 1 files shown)


craigraw:
Interesting - haven't had a chance to look into this yet. Is there a reason to make the delay 5 seconds rather than 2?

PrinceOfEgypt:
I tested it on windows, Mac, and Linux:

- The Mac would never manifest the issue even with no delay
- Ubuntu 23.04 worked every time with the 2 second delay
- Windows 10 still had the error once even with the 2 second delay, but could not reproduce the issue when I increased it to 5

craigraw:
Interesting - typical of Windows to be worst. I'd prefer to get the request out sooner than 5 seconds if possible. What about 
```
ratesService.setDelay(Duration.seconds(org.controlsfx.tools.Platform.getCurrent() == org.controlsfx.tools.Platform.WINDOWS ? 5 : 2));
```

(with appropriate constants named and declared as you have done)

PrinceOfEgypt:
Ok, in the interest of clarity I added a switch statement with a default value:

```
        switch(org.controlsfx.tools.Platform.getCurrent())
        {
            case WINDOWS:
                //Windows needs more time (5 Seconds)
                ratesService.setDelay(Duration.seconds(RATES_STARTUP_DELAY_SECS_WINDOWS));
                break;
            case OSX:
            case UNIX:
            case UNKNOWN:
            default:
                //Other platforms seem ok with 2 Seconds
                ratesService.setDelay(Duration.seconds(RATES_STARTUP_DELAY_SECS_DEFAULT));
                break;
        }
```

craigraw:
Thanks, appreciate the cross-platform testing on this!

craigraw:
I have edited it a bit for brevity. I'm a fan of keeping things terse :) Appreciate the work nonetheless.

PrinceOfEgypt:
> I have edited it a bit for brevity. I'm a fan of keeping things terse :) Appreciate the work nonetheless.

No worries.. everyone had their own style :)

PR #1369, opened 2024-04-10
Whirlpool 1.0.6

zeroleak:
- preserve single ExtLibJConfig instance on all services to make onion switch easy


craigraw:
Thanks!

PR #1437, opened 2024-06-08
Add SingleMainWindow=true to Sparrow.desktop

### LINKED ISSUE #1436: .desktop should set Single Window
Hello Sparrow

I noticed the .desktop file included in the linux packages displays "New Window" in the context menu of right clicking on the Icon in Gnome but Sparrow is a single window application so this is wrong and does nothing useful if clicked.

Current Incorrect Behavior:
image

image

This can be prevented by adding this to key to the .desktop file:
```
SingleMainWindow=true
```

Desired Behavior after adding the above:
image

image

BenWestgate:
This prevents desktop environments from displaying "New Window" as one of the right click actions in the side bar and application list when Sparrow is already running.

Closes #1436.


ArkFunds (approved):

craigraw:
I've noticed that this flag doesn't seem to do anything in Ubuntu 20.04 and Debian 11.3 - the `New Window` menu command appears regardless, but doesn't do anything - as is the case presently. Newer versions of these distros do seem to have support, although sometimes a restart is necessary.

It's still a net benefit though, so merging. Thanks!

PR #1500, opened 2024-09-17
Multilanguage

anurag-lint:
I have implemented the multi-language functionality. The process works as follows:

### Change from menu
From the menu option `View --> Language`, the language can be selected. Once the language is chosen, the application will restart in the same way as when clicking on `Restart in` to switch networks.

image

image

### Change from welcomeDialog
- In the WelcomeDialog, which opens when launching Sparrow after the first installation (or the first time it is run on some of the test networks), the first step of the wizard will also allow the selection of the language. If the WelcomeDialog is opened later from the `Tools` menu, the language change will no longer be allowed there, and it must be done from the menu.

Captura de pantalla 2024-09-17 a las 8 16 49
image

Currently, I have translated the WelcomeDialog, the Transactions tab and the Send tab, pending validation of this proposed solution.

The way to add new languages will be to add a new element to the `Languages.java` enum and include the `.properties` file for the new language, which will contain the translations. It is important to add the message for the name of the new language (for example, `language.fr` for French) to both the new file and the existing ones.

### DIFF
--- a/src/main/java/com/sparrowwallet/sparrow/AppController.java
+++ b/src/main/java/com/sparrowwallet/sparrow/AppController.java
@@ -141,7 +141,10 @@ public class AppController implements Initializable {
     private ToggleGroup theme;
 
     @FXML
-    private ToggleGroup language;
+    private ToggleGroup languages;
+
+    @FXML
+    private Menu languagesMenu;
 
     @FXML
     private CheckMenuItem openWalletsInNewWindows;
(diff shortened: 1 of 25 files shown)


ThauanAmorim:
Still draft?

hiciefte:
Is multi-language support something the project would consider taking on? The approach here looks right to me — the usual JavaFX pattern, a `Languages.java` enum plus one `.properties` per locale.

The thing that usually stalls wallet i18n isn't the first translation, it's the upkeep once English strings start changing across releases. I help maintain Bisq's localization tooling ([localize-pipeline](https://github.com/bisq-network/localize-pipeline)), which keeps JavaFX `.properties` locales in sync by opening a PR for review whenever the source changes — opt-in, never auto-merged, and it can run against a local model so nothing leaves your machine. It only edits the `.properties` data files, so it doesn't touch build logic or reproducibility.

I'm not asking anyone to merge this — mostly trying to understand whether i18n is something you'd take on at all, and if so, what the bar would be for a translation in a wallet, where a bad string in the send flow or a fee warning actually costs money.

PR #1506, opened 2024-09-22
Add input labels into TransactionDiagram along with transaction notes

canselcik:
Will add screenshots later if needed. I don't imagine any downsides to it and it feels like something that's missing.

### DIFF
--- a/src/main/java/com/sparrowwallet/sparrow/control/TransactionDiagram.java
+++ b/src/main/java/com/sparrowwallet/sparrow/control/TransactionDiagram.java
@@ -429,11 +429,18 @@ public class TransactionDiagram extends GridPane {
 
                 Tooltip tooltip = new Tooltip();
                 Long inputValue = null;
+                String inputLabel = input.getLabel();
                 if(walletNode != null) {
                     inputValue = input.getValue();
                     Wallet nodeWallet = walletNode.getWallet();
-                    tooltip.setText("Spending " + getSatsValue(inputValue) + " sats from " + (isFinal() ? nodeWallet.getFullDisplayName() : (nodeWallet.isNested() ? nodeWallet.getDisplayName() : "")) + " " + walletNode + "\n" +
-                            input.getHashAsString() + ":" + input.getIndex() + "\n" + walletNode.getAddress());
+                    StringJoiner joiner = new StringJoiner("\n");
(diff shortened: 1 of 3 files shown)


craigraw:
Not sure I follow. 
- The input label is displayed on the input already, displaying it on the tooltip as well is redundant (see `getInputDescription` method)
- The name of the transaction is already displayed as the tab name.

canselcik:
@craigraw Thank you for the reply, I appreciate it.

> The input label is displayed on the input already, displaying it on the tooltip as well is redundant (see getInputDescription method)
The input label is displayed in a shortened way on the left side. If I have a long note in the label, it might as well be not there.

> The name of the transaction is already displayed as the tab name.
This hasn't been my experience, I see just the first six characters of the hash in brackets in the tab title. 

Here is the screenshot I promised earlier. Let me know what you think.
example

craigraw:
> The input label is displayed in a shortened way on the left side. If I have a long note in the label, it might as well be not there.

Ok, that's reasonable. I remember someone else mentioning this as well. Can we keep the same formatting though - no space after the `if`. I'm aware it's non-standard :)

> This hasn't been my experience, I see just the first six characters of the hash in brackets in the tab title.

This occurs only if there is no name. For reference, see https://github.com/sparrowwallet/sparrow/blob/e1dab3a48e796e00c2c01d6f3cf8d023fec15d61/src/main/java/com/sparrowwallet/sparrow/AppController.java#L1968-L1970

But I understand what you mean - when opening transactions with existing labels, the tab name did not take on the label. This has been fixed in 1bbc586c.

canselcik:
No worries, happy to keep the `if` the non-standard way you like. :)

I appreciate the quick fix with https://github.com/sparrowwallet/sparrow/commit/1bbc586cd616e46c6ebd9615fb0e616ae44fb98c, I will also rebase this commit and revert the reformatted if.

Another thing that I think would be great to have in Sparrow would be an easy to navigate linear utxo history.
It can even be as simple as a ListView but being able to see all the transactions a utxo has been involved with, with their dates and transaction labels would be very helpful.

I would try to add this myself as well but I am not familiar with the codebase enough to put together a proof of concept real quick.

craigraw:
Just following up here? - happy to just make the changes myself, no problem. Mostly just reverting HeadersController and headers.fxml.

craigraw:
Added in e92d0f9b.

PR #1537, opened 2024-11-15
Close wallet name dialog with ESC

ottosch:


craigraw:
Thanks - normally I would suggest using `AppServices::onEscapePressed`, but in this case the textfield is eating the key pressed event.

PR #1591, opened 2025-01-14
Patch https://github.com/Toporin/SatochipApplet/issues/15

Toporin:
Null exception can be thrown when signing a multisig transaction from a Sparrow wallet reconstructed from a Bitcoin descriptor. This happens when the user did not configure any keystore with the corresponding Satochip card ('import' button). In this case, the 'fullpath' derivation path remains undefined, leading to the exception.


craigraw:
Doesn't this change make the (possibly incorrect) assumption that the first keystore in the wallet has the same derivation path as the Satochip keystore, whichever one it may be?

craigraw:
This is a tricky problem. One approach that would strictly better is to only default to the first keystore if the Satochip keystore cannot be determined (from the wallet model as currently, or some other approach).

craigraw:
LGTM

PR #1643, opened 2025-02-20
Automatically apply system theme on startup

triosi:
I’ve added a "System Default" theme option that reads the system theme and applies it to Sparrow (switching between Light and Dark) at startup. This change ensures Sparrow launches with the same theme as the user’s system preference.

**Changes Made:**
- Added a `detectSystemTheme` method to identify the system theme.
- Implemented `isMacDarkMode` to check the `AppleInterfaceStyle` setting on macOS.
- Modified the startup logic to apply the system theme when "System Default" is selected.

**Limitations:**
- Currently, this feature only works on macOS. I don’t have test environments for Linux or Windows, so I haven’t implemented theme detection for those platforms yet.
- I didn’t add dynamic theme switching (i.e., updating the app’s theme when the system theme changes while the app is running). Implementing this might require polling the system theme continuously, which could be resource-intensive. I’m unsure if the added complexity is worth it and would appreciate feedback on this.
- I wasn’t able to apply dark mode to the welcome screen during its initial launch, so it remains in light mode even when the system theme is dark.

**Future Steps:**
- If this PR is deemed useful, I’m willing to extend support to Windows and Linux. I can write the necessary code and set up virtual environments to test them.


craigraw:
Thanks for the suggestion. A little nitpick that the `OsType` class should be used instead of `System.getProperty("os.name")`. Another is that `isMacDarkMode` should catch `Throwable`, as any failure will prevent app startup.

On a broader scope, I've taken a look at https://github.com/Dansoftowner/jSystemThemeDetector and it seems there's more than one approach. I'm not particularly keen to add all the dependencies that project has for such a simple function, but it's worth considering the techniques used there.

I think I'd want to put in more work into the dark theme before merging this. Specifically, there are images that require SVG alternatives so they can be properly colour inverted in dark mode.

triosi:
Hi @craigraw,

I’ve made the two changes you requested.

Regarding the library you suggested, do you think it is actually necessary for the required functionality? I tested my current implementation on macOS, and it seems to provide the necessary behavior without the additional dependency.

Also, are you still planning to work on the dark theme itself? I think the “Auto” functionality and having a fully polished dark theme are somewhat separate concerns. In my view, we could enable the Auto option now and improve the dark theme independently.

As I see it, the only thing needed for this PR to be complete is extending this functionality to Windows and Linux. However, I currently only have access to macOS, so I wouldn’t be able to properly test the implementation on those platforms. I could look into setting up VMs if needed, but I wanted to share these thoughts with you first.

I’d be happy to continue working on this if you think this is a good direction for the PR.

Thanks!

craigraw:
Thanks for this, and for pushing on the "Auto" option. I agree it didn't need to wait on the dark theme work, which is now done anyway.

I've implemented it in a573f22a, but differently from this PR. I'd rather not start an external process to find the system theme. JavaFX 22 added `Platform.getPreferences()`, and its `colorScheme` property reads the OS setting on macOS, Windows and Linux. Because the property can be listened to, Sparrow can follow a change while it's running, with no polling. So a single approach covers all three platforms, and nobody needs to set up VMs to test detection separately.

A few other changes compared with this PR:

- The theme is resolved in one place, and every place that previously checked for Dark now uses it. With `SYSTEM` stored in config, those checks would otherwise have treated the theme as light, leaving dialogs, icons and SVG images light while the main window was dark.
- A system change also updates open dialogs. Most follow their owner window. Send to Many and Search Wallets have no owner, so they're updated directly.
- System is now the default for new installs. Existing configs keep their saved Light or Dark setting.

On Linux the colour scheme is inferred from the GTK theme's colours, so a "prefer dark" desktop setting with a light GTK theme may still show as light.

Closing this in favour of that commit. Thanks again for getting it started.

triosi:
Thank you for your dedication and the excellent solutions you developed for the issues, @craigraw. Bitcoin is growing more practical because of your efforts.

PR #1649, opened 2025-02-23
GHA package.yml: fail-fast:false

msgilligan:
With this setting, if a test on one matrix.os fails, the other tests will not be cancelled.


craigraw:
Again, no thanks - I'd rather have the other builds fail immediately, since the whole thing will need to be redone anyway after the failing build is fixed.

msgilligan:
>  I'd rather have the other builds fail immediately, since the whole thing will need to be redone anyway after the failing build is fixed.

I can see that. But on the other hand, I find it can save time (especially with long build/tests jobs) to know all the platforms that are failing before the next commit.

PR #1656, opened 2025-02-25
GHA package.yaml: Add explicit job name

msgilligan:
* ~~Change job label from `build` to `jpackage`~~
* Add a `name` attribute setting name with matrix.os

This matches the format of `package-nix.yaml` in PR #1648 to keep them similar assuming it is merged as is.

### DIFF
--- a/.github/workflows/package.yaml
+++ b/.github/workflows/package.yaml
@@ -7,6 +7,7 @@ permissions:
 
 jobs:
   build:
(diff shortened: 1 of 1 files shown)


msgilligan:
> Can I suggest instead...

@craigraw Done. Rebased and force-pushed.

kmk142789 (approved):

PR #1750, opened 2025-06-07
Replace JNI-based ZBar wrapper with jzbar (Java FFM)

doblon8:
This PR replaces the current JNI-based ZBar integration with [jzbar](https://github.com/doblon8/jzbar), a minimal Java FFM wrapper I wrote for the subset of ZBar functionality currently used by Sparrow.

Note: currently, `jzbar` looks for a system-installed `zbar` shared library:
```java
SymbolLookup.libraryLookup(System.mapLibraryName("zbar"), LIBRARY_ARENA)
    .or(SymbolLookup.loaderLookup())
    .or(Linker.nativeLinker().defaultLookup());
```
It does not use the native libraries bundled in Sparrow’s `src/main/resources/native`. Since this is just a draft, I’d like to hear your thoughts on whether the final version should use those instead.

I’ve verified it works by successfully importing a wallet from a Blockstream Jade device and signing a transaction with it, though further testing would still be necessary.

Let me know if you consider this an improvement. If so, I’m happy to refine the PR further and include any changes you think are necessary.


doblon8:
Forgot to mention that, since `jzbar` isn’t published to any public Maven repository yet, you’ll need to build and install it locally first using Maven:
```bash
mvn clean install
```
That installs the library to your local Maven repository (`~/.m2/repository`), so Gradle can resolve it.

craigraw:
Thanks, looks like a good start! I've been working to get native binaries out of the Sparrow codebase and into libraries, so this is useful. Appreciate the work.

In terms of where the native binaries should go, my preference would be in the jzbar JAR itself. This makes the library an atomic unit for upgrades etc. That probably means extracting them to a temporary location each time before loading them, but on the other hand this is a tried and tested technique many libraries use. The alternative is for me to place them in the installed Sparrow application folder tree - also possible, but more complicated to maintain. 

On the PR itself, it probably makes more sense to put ZBar.java into the `com.sparrowwallet.sparrow.io` package. Otherwise seems straightforward.

Including @msgilligan on this, since he has been working on an FFM wrapper for libsecp256k1 and might have some additional thoughts.

doblon8:
> In terms of where the native binaries should go, my preference would be in the jzbar JAR itself. This makes the library an atomic unit for upgrades etc. That probably means extracting them to a temporary location each time before loading them, but on the other hand this is a tried and tested technique many libraries use. The alternative is for me to place them in the installed Sparrow application folder tree - also possible, but more complicated to maintain.

I've included the native binaries inside the `jzbar` JAR as you suggested. I also added builds for multiple OSes in the GitHub Actions workflow, and they appear to work fine.
> 
> On the PR itself, it probably makes more sense to put ZBar.java into the `com.sparrowwallet.sparrow.io` package. Otherwise seems straightforward.

Done as well.

Thanks for the suggestions, and happy to make further changes if needed.

build.gradle
```diff
@@ -23,6 +23,7 @@ group 'com.sparrowwallet'
 version '2.2.3'
 
 repositories {
+    mavenLocal()
```
doblon8, line 26: I'm still working on setting up a proper namespace to publish this library to Maven Central. Once that's done, I plan to remove this `mavenLocal()` line.

msgilligan:
> Including @msgilligan on this, since he has been working on an FFM wrapper for libsecp256k1 and might have some additional thoughts.

Nice to see this work! One cool feature I've investigated and briefly tested is that (with Java 25) you can natively compile apps that are using FFM and have natively-compiled Java code which calls the native library through FFM. Unfortunately, my experiences with natively-compiled JavaFX apps have not (yet) convinced me that an app like Sparrow could be natively compiled.

However, I think it is a use-case that FFM wrapper libraries should try to support.

craigraw:
> Thanks for the suggestions, and happy to make further changes if needed.

Thanks, changes look good. The only other change I would suggest is to add a `module-info.java` file so the lib is modular and doesn't need the additional config and modification in Sparrow's build.gradle `extraJavaModuleInfo`.

doblon8:
> Thanks, changes look good. The only other change I would suggest is to add a `module-info.java` file so the lib is modular and doesn't need the additional config and modification in Sparrow's build.gradle `extraJavaModuleInfo`.

Done. Added `module-info.java` to `jzbar`.

craigraw:
Thanks - I'll wait until it's Maven Central before making the update. No rush, there won't be another Sparrow release for awhile.

doblon8:
I’ve published `jzbar` to Maven Central and updated the project to depend on it instead of the local version.

craigraw:
Thanks! I've made a minor followup change in 3c5fa58a to suppress JVM warnings.

PR #1775, opened 2025-07-01
Do not change coin label unit on right click

Radiokot:
Hey Craig,

This is a tiny fix preventing Bitcoin unit change in labels such as "Balance" and "Mempool" when you right-click them to copy the value.

Thank you for your work on Sparrow.


craigraw:
Thanks!

Issue #1771, opened 2025-06-23
Enable TX Creation under 1 sat/VB


RobertClarke:
BTC is more expensive than ever, the mempool is empty. Enabling sub 1sat/VB TX generation from within Sparrow will allow the growing [percentage of hashpower (currently estimated @ 40% of the network)](https://x.com/mononautical/status/1936234974736466305) to mine these transactions, saving users 90% or more on their transactions which are significant, especially on UTXO consolidation transactions (a use-case that Sparrow is often best at coordinating).

Full nodes with config minrelaytxfee=0.0001 will be able to relay transactions @ 0.1 sat/VB [and they are being consistently mined](https://x.com/peterktodd/status/1935709056695775583).

Raimo33:
Anyone wants to tackle this? It shouldn't be hard to implement. And a big UX improvement.

francitoshi:
the 0.1 sat/vbyte is the new 1 sat/vbyte, as Bitcoin goes up in price 1 sat/vbyte is going to much.

craigraw:
I've implemented support for `minrelaytxfee` in 9dcf2107. The server you are connecting to must change `minrelaytxfee` in bitcoin.conf - this value will be read by Sparrow either directly from Bitcoin Core or indirectly via the Electrum server that connects to that node. To override this value, close Sparrow, edit the Sparrow config file in Sparrow home and change the value of `minRelayFeeRate` to a value >= `0.0` and < `1.0`. Note that this is considered advanced usage - if the server you are connected to doesn't support relaying transactions at the chosen fee rate, it will return an error on broadcast.

Since this appears to be a much anticipated feature, I've created a pre-release build for testing: 
https://github.com/craigraw/beta/releases/tag/2.2.4-support-minrelaytxfee

Please provide feedback here.

francitoshi:
Tested, it works ok. 
1) if you connect to your own node with minrelaytxfee configured, it works.
2) if you connect to your own node minRelayFeeRate configured in Sparrow it works and has priority over the node minrelaytxfee if configured.
3) if you connect to a public node with minRelayFeeRate configured in Sparrow it works.

hazrulnizam:
Tested the pre-release by connecting to own private Electrum server.

Works great! Sparrow automatically detected the minrelayfee settings and allowed me to create subsat txs.

RobertClarke:
Working well for me as well, thanks for the add @craigraw

craigraw:
[...]

PR #1791, opened 2025-07-15
Support for fee rate below 1sat/vb

ThauanAmorim:


jaonoctus (approved):

happyluckbox:
Great idea, would love to have the option to save on fees when possible!

craigraw:
Thanks for exploring the UI on this. I have been considering it myself, and have some alternative approaches in mind.

That said, one cannot simply change the minimum fee rate, since the node the wallet is connected to will not accept transactions below `minrelaytxfee` for broadcast - the value must first be determined from the server.

> the message informs that transactions can be purged from mempool.

Transactions can always be purged from the mempool - the question is whether they are relayed to mempools with default policy settings.

> I've fixed a bug. If an invalid number was typed in the fee text box, the create button was not disabled.

Can you provide steps to reproduce this? I can't on my side.

ThauanAmorim:
> Thanks for exploring the UI on this. I have been considering it myself, and have some alternative approaches in mind.
> 
> That said, one cannot simply change the minimum fee rate, since the node the wallet is connected to will not accept transactions below `minrelaytxfee` for broadcast - the value must first be determined from the server.
> 
> > the message informs that transactions can be purged from mempool.
> 
> Transactions can always be purged from the mempool - the question is whether they are relayed to mempools with default policy settings.

Nodes supporting fee below 1 sat/vb are becoming increasingly common.
Even when connected to the node itself or to a third-party node with the reduced minimum fee, it was not possible to create transactions with the reduced rate on sparrow.

My idea was to simply add the support and leave the yellow field to warn of a possible failure.

I also thought about displaying the low fee only if connected to a node that supports them, but this will cause inconsistencies in usage.
some times the minimum fee will be available and other times it won't be.

Do you have any other ideas on how to improve? Not having support for the minimum fee seems to me to be the worst

> > I've fixed a bug. If an invalid number was typed in the fee text box, the create button was not disabled.
> 
> Can you provide steps to reproduce this? I can't on my side.

On "send" screen, simply fill in all the fields with valid values and then type 0 or just remove the fee value. The creation field will still be available.

ThauanAmorim:
I've updated the description

ThauanAmorim:
Updated message.
Old: Transactions at this fee rate can be purged from the default sized mempool
new: Transactions at this fee rate can be purged from mempool
image

craigraw:
Thanks for your work on this. I want to respect your time - I won't be accepting this PR as I am taking a slightly different approach as I'll discuss below. But I appreciate the initiative, and the ideas. 

> I also thought about displaying the low fee only if connected to a node that supports them, but this will cause inconsistencies in usage.
some times the minimum fee will be available and other times it won't be.

The problem with not respecting the server's configuration is that users will create transactions that the server does not accept for broadcast, causing a lot of confusion and user support requests. I intend to allow advanced users to override the server configuration in Sparrow (for example, to broadcast manually on mempool.space), but for regular usage it must be respected.

> On "send" screen, simply fill in all the fields with valid values and then type 0 or just remove the fee value. The creation field will still be available.

Thanks. I've fixed this on my side.

I'll update #1771 soon with my progress on sub 1 sat/vb support.

epcgrs (commented):
LGTM! 

This PR refactors the fee rate selection logic to use a more precise List<Double> for fee rates instead of List<Long>, and updates the relevant UI components, validation, and tooltips accordingly.

ThauanAmorim:
> Thanks for your work on this. I want to respect your time - I won't be accepting this PR as I am taking a slightly different approach as I'll discuss below. But I appreciate the initiative, and the ideas.
> 
> > I also thought about displaying the low fee only if connected to a node that supports them, but this will cause inconsistencies in usage.
> > some times the minimum fee will be available and other times it won't be.
> 
> The problem with not respecting the server's configuration is that users will create transactions that the server does not accept for broadcast, causing a lot of confusion and user support requests. I intend to allow advanced users to override the server configuration in Sparrow (for example, to broadcast manually on mempool.space), but for regular usage it must be respected.
> 
> > On "send" screen, simply fill in all the fields with valid values and then type 0 or just remove the fee value. The creation field will still be available.
> 
> Thanks. I've fixed this on my side.
> 
> I'll update #1771 soon with my progress on sub 1 sat/vb support.

What do you think of the idea of releasing it only to full nodes? I can recover the value of the minimum fee and use it to update the range.

I don't see how the code could be much different from that. The big problem is transaction propagation, but if the node has the minimum fee configured, I would release the range with the lower rates.

`minimumRelayFeeRate = Math.max(event.getMinimumRelayFeeRate(), Transaction.DEFAULT_MIN_RELAY_FEE);`
Currently it always recovers the maximum value to keep it above 1 sat/vb. Changing it to retrieve the lowest value and based on the number define whether the rates below the minimum can be used or not would already solve the problem for more novice users.

epcgrs:
Hi @craigraw, first of all, thank you for your incredible work on Sparrow. It's one of the most powerful wallets in the ecosystem.
I believe this PR has real value for advanced users running full nodes with custom `minrelaytxfee` settings.
Would you consider allowing this feature behind an optional advanced flag — something like:

```[x] Enable sub 1 sat/vB fees (requires full node support)```

This way, regular users stay protected, and advanced users retain full control in low-congestion scenarios where very low-fee txs are actually valid and relayed by their own nodes. Having this available in Sparrow would be extremely useful for those experimenting with custom relay policies or constrained environments.

Thanks again for the amazing work. Looking forward to your thoughts!

ThauanAmorim:
What do you think of this version? 
I modified the minimum rate calculation to save the lowest rate found. If it is 0.01, it releases the list with all available rates; if it is higher than 0.01, only the standard rates will be available.

When connecting to at least one node that supports low rate
image

When no node supports low rate
image

ThauanAmorim:
https://github.com/sparrowwallet/sparrow/pull/1792
new PR. If you don't like it, you can close it.

PR #1792, opened 2025-07-16
Low fee

ThauanAmorim:
Some images are not up to date, so the recent blocks are not displayed, but the behavior has remained the same

First option: I didn't like it because there are too many zeros.
image

Current version. It broke the pattern, but it's cleaner.
image

When the range is moved all the way to the right, the smallest values are removed.
image

When the minimum fee is reached, everything stays the same.
image

When below the minimum fee now the field turns yellow as an alert and the message informs that transactions can be purged from mempool.
image

When 0.01, now it's “Try Then Replace”.
image

If the fee is zero, it remains red to report an error.
image

I've fixed a bug. If an invalid number was typed in the fee text box, the create button was not disabled.

What do you think of this version? 
I modified the minimum rate calculation to save the lowest rate found. If it is 0.01, it releases the list with all available rates; if it is higher than 0.01, only the standard rates will be available.

When connecting to at least one node that supports low rate
image

When no node supports low rate
image

Old PR: https://github.com/sparrowwallet/sparrow/pull/1791

### DIFF
--- a/src/main/java/com/sparrowwallet/sparrow/AppServices.java
+++ b/src/main/java/com/sparrowwallet/sparrow/AppServices.java
@@ -16,6 +16,7 @@ import com.sparrowwallet.drongo.wallet.*;
 import com.sparrowwallet.sparrow.control.DialogImage;
 import com.sparrowwallet.sparrow.control.WalletPasswordDialog;
 import com.sparrowwallet.sparrow.glyphfont.FontAwesome5;
+import com.sparrowwallet.sparrow.net.Auth47;
 import com.sparrowwallet.drongo.protocol.BlockHeader;
 import com.sparrowwallet.drongo.protocol.ScriptType;
 import com.sparrowwallet.drongo.protocol.Transaction;


craigraw:
> What do you think of this version? I modified the minimum rate calculation to save the lowest rate found. If it is 0.01, it releases the list with all available rates; if it is higher than 0.01, only the standard rates will be available.

This is better, and more similar to my approach. But there's a great deal more to consider, and I've already completed that work - it's in the testing phase, and I'll commit it soon when I'm satisfied all the edge cases have been taken care of.

Put another way, there's no need to work on this - it's coming.

ThauanAmorim:
> > What do you think of this version? I modified the minimum rate calculation to save the lowest rate found. If it is 0.01, it releases the list with all available rates; if it is higher than 0.01, only the standard rates will be available.
> 
> This is better, and more similar to my approach. But there's a great deal more to consider, and I've already completed that work - it's in the testing phase, and I'll commit it soon when I'm satisfied all the edge cases have been taken care of.
> 
> Put another way, there's no need to work on this - it's coming.

got it

douglaz:
@craigraw do you have a PR/branch so users can test it?

craigraw:
@douglaz There is a pre-release build listed on the original issue #1771.

PR #1819, opened 2025-08-28
add Bitkey support

harrisonfriia:
- Adds support for importing Bitkey multisig wallets. 
- Handles Bitkey's non-standard m/84'/0'/0' + P2WSH derivation paths.


craigraw:
Thanks for the PR. I think this needs some motivation though. Since this is just creating a watch only wallet from the BitKey output descriptor, why not just use File > Import Wallet > Output Descriptor? This has been supported since https://github.com/sparrowwallet/sparrow/commit/799cac7b1f75049ddf834a3752138ba1f1da4e64.

harrisonfriia:
> Thanks for the PR. I think this needs some motivation though. Since this is just creating a watch only wallet from the BitKey output descriptor, why not just use File > Import Wallet > Output Descriptor? This has been supported since [799cac7](https://github.com/sparrowwallet/sparrow/commit/799cac7b1f75049ddf834a3752138ba1f1da4e64).

Thanks for the heads up - you’re right, this does overlap with that. I started this before that landed and missed the update. I’ll go ahead and close this PR.

Issue #476, opened 2022-03-21
Feature Request: Drag tabs to change order


statusquont:
A nice-to-have: the ability to drag tabs within the UI to re-order them. A use case might be that I have multiple wallets open working with several transaction details tabs, and I want to mentally group transactions with their corresponding wallet by creating a visual clue, such as:

[wallet a] [wallet a tx 1] [wallet a tx 2] [wallet b] [wallet b tx 1] [wallet c]

Right now it seems all tx tabs just get appended to the end of the tab list, which would be ok if I could drag them into the position I want.

craigraw:
Draggable tabs is a feature I would like to add, but there are issues in the underlying UI framework I'm not comfortable with. Specifically, enabling dragging can cause the selected tab to jump to the right when clicking on it, a behaviour I find sufficiently disconcerting that I have disabled draggable tabs. I need to report this upstream so it can hopefully be improved. Example below:

![draggable](https://user-images.githubusercontent.com/862166/159428470-1bdf9761-e8df-453a-9a8c-18686ceae4f6.gif)

craigraw:
Asked on the openjfx mailing list about this problem, unfortunately without response: https://mail.openjdk.org/pipermail/openjfx-dev/2022-August/035582.html

Although it's not draggable tabs as such, bacbdb84 adds support for reordering tabs via the tab context menu.

<img width="249" alt="Screenshot 2022-09-06 at 12 44 26" src="https://user-images.githubusercontent.com/862166/188616125-753cc9bd-438f-4da8-b2f5-80a46871af56.png">

Giszmo:
The context menu indeed is helpful dragging would still be very welcome.

The context menu here does not open randomly. I right-click on a tab many times and on another one many times and only when trying the third tab it works and then it works for others, too and I think when switching tabs this not working context menu re-emerges.

craigraw:
@Giszmo it seems like the context menu problem you mention is a separate issue?

Giszmo:
@craigraw the context menu issue is a glitch and glitches in wallet software always make me nervous but yes, it's only remotely related to this issue here. Sorry I'm not setup to document the glitch well in its own issue.

m-o-t-u:
It would also be very helpful if I could move the tabs to the right or left using keyboard shortcuts. #1836 pr created

PR #1836, opened 2025-09-23
Move tabs by keyboard

m-o-t-u:
Adds the ability to move the active tab using the keyboard shortcut CTRL+Shift+LEFT / CTRL+Shift+RIGHT #476


craigraw:
This is already supported by the Ctrl+PageUp/PageDown keys on Windows and Linux (using standard JavaFX bindings). Is there a need for additional keys?

m-o-t-u:
Ctrl+PageUp/PageDown only changes the current tab (focus), but I can't use it to rearrange the tabs.
I work with about 10 tabs, and sorting them is a bit tedious with “right-click - move left/right.”

craigraw:
Fair enough.

I found this was simpler to implement by adding accelerators to the current tab context menu items, which I've done in 0f4c36b3.

Thanks for the suggestion though, and appreciate the PR!
Screenshot 2025-10-13 at 14 24 44

Issue #1525, opened 2024-10-25
Feature request: Support BIP93 (codex32)


ikmckenz:
[BIP-0093](https://github.com/bitcoin/bips/blob/master/bip-0093.mediawiki) aka [Codex32](https://www.secretcodex32.com/) is a Shamir's secret sharing scheme for BIP32 seeds. The scheme has many advantages over BIP39 words, and enables (among other things) the ability to hand verify checksums to verify data integrity, and is thus interesting for very long term backup of Shamir shares, even when secrets are stored in a software wallet like Sparrow. 
Minimal support for BIP93 would include the ability to load the single secret share in the codex32 format, and then generate addresses and sign transactions from there. More complete support would include the ability to calculate the secret and import from k-of-n shares, and to compute shares from a newly generated seed.

ikmckenz:
Made some initial progress towards this goal in the upstream drongo library:

First I put in some unit tests to make sure I didn't break anything:
1. sparrowwallet/drongo/pull/32
2. sparrowwallet/drongo/pull/33
3. sparrowwallet/drongo/pull/34

Now I've released a PR for my (basic) Codex32 implementation:
1. sparrowwallet/drongo/pull/35

Assuming that my implementation gets merged into the library without major changes, I think the next step is to start to work on support in the sparrow repository.

ikmckenz:
Initial support is now merged in the sparrow wallet: #1892

This means importing the single secret share in the codex32 format and using it as a wallet now works. 

What is still missing is importing from k-of-n (non-secret) shares, error correction, and computing shares from a newly generated seed. It may be good to wait until some draft upstream recommendations for wallet generation are resolved before implementing the generation part: BlockstreamResearch/codex32/pull/71

PR #1892, opened 2025-11-30
Add Codex32 wallet import

ikmckenz:
Draft PR to add codex32 wallet import. It seems to be working as is, but needs improvement before merging. I've never worked with javafx before so I'm probably doing some things wrong. Once eventually merged it should finish the "initial implementation" part of #1525. 

Prerequisite is sparrowwallet/drongo/pull/37. 

Another prerequisite would be adding a logo to `src/main/resources/image/walletmodel`, I was trying to make a logo but I'm pretty bad at it so I would probably need help here :)

Couple improvements to make: `getScriptTypeEntry()` and `getWallet()` are mostly copied from `MnemonicWalletKeystoreImportPane.java`, I was thinking of inserting a new abstraction layer that extends `TitledDescriptionPane` to contain those functions, then `MnemonicKeystorePane` could extend from that, but it kinda got complicated. Open to suggestion how to improve code reuse here.

### DIFF
--- a/src/main/java/com/sparrowwallet/sparrow/control/CodexKeystoreImportPane.java
+++ b/src/main/java/com/sparrowwallet/sparrow/control/CodexKeystoreImportPane.java
@@ -1,0 +1,251 @@
+package com.sparrowwallet.sparrow.control;
+
+import com.sparrowwallet.drongo.KeyDerivation;
+import com.sparrowwallet.drongo.crypto.ChildNumber;
+import com.sparrowwallet.drongo.protocol.ScriptType;
+import com.sparrowwallet.drongo.wallet.Keystore;
+import com.sparrowwallet.drongo.wallet.Wallet;
+import com.sparrowwallet.sparrow.EventManager;
+import com.sparrowwallet.sparrow.event.KeystoreImportEvent;
+import com.sparrowwallet.sparrow.glyphfont.GlyphUtils;
+import com.sparrowwallet.sparrow.io.ImportException;
+import com.sparrowwallet.sparrow.io.KeystoreCodexImport;
+import javafx.application.Platform;
+import javafx.beans.property.SimpleStringProperty;
+import javafx.beans.value.ChangeListener;
+import javafx.geometry.Insets;
+import javafx.geometry.Pos;
+import javafx.scene.Node;
+import javafx.scene.control.Button;
+import javafx.scene.control.ContentDisplay;
+import javafx.scene.control.Control;
+import javafx.scene.control.Label;
+import javafx.scene.control.MenuItem;
+import javafx.scene.control.SplitMenuButton;
+import javafx.scene.control.TextField;
+import javafx.scene.control.Tooltip;
+import javafx.scene.layout.AnchorPane;
+import javafx.scene.layout.HBox;
+import javafx.scene.layout.Priority;
+import javafx.scene.layout.VBox;
+import org.controlsfx.validation.ValidationResult;
+import org.controlsfx.validation.ValidationSupport;
+import org.controlsfx.validation.Validator;
+import org.controlsfx.validation.decoration.StyleClassValidationDecoration;
+
+import java.util.List;
+
+public class CodexKeystoreImportPane extends TitledDescriptionPane {
+    protected final Wallet wallet;
+    private final KeystoreCodexImport importer;
+    private final KeyDerivation defaultDerivation;
+
+    private SplitMenuButton importButton;
+
+    private Button enterCodexButton;
+    private Button calculateButton;
+
+    protected Label validLabel;
+    protected Label invalidLabel;
+
+    protected final SimpleStringProperty secretShareProperty = new SimpleStringProperty("");
+
+    public CodexKeystoreImportPane(Wallet wallet, KeystoreCodexImport importer, KeyDerivation defaultDerivation) {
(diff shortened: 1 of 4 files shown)


craigraw:
Some general notes:
1. We actually don't need https://github.com/sparrowwallet/drongo/pull/37 - `WalletModel.SEED` is sufficient, see `Bip32.java`.
2. The wallet import dialog might be a place to add this in future, but it's not the right place to start I think. I have not added the Bip32 xprv import or the SLIP39 import here as they are both rarely used. Consider rather referencing the `TitledDescriptionPane` subclass off `MnemonicShareKeystoreImportPane` which is instantiated in `SwController.java` and added to the list of options when creating a wallet using the File > New Wallet flow. This is the more declarative approach to creating a wallet and a better place to add new seed import formats. The import UI pane would then be instantiated as 
```
public CodexKeystoreImportPane(Wallet wallet, KeystoreCodexImport importer, KeyDerivation defaultDerivation) {
   //wallet contains a reference to the scriptType etc
}
```

ikmckenz:
Yep, that seems right to me. I probably should have copied the SLIP39 stuff in the first place. 

One prerequisite I'm pretty sure is actually needed (and originally forgot to commit) is exporting the bip93 module in drongo: sparrowwallet/drongo/pull/38

I've updated the pull request and it is now importing via the File > New Wallet flow, and works similar to `MnemonicShareKeystoreImportPane`. Seems to be working to me.

src/main/java/com/sparrowwallet/sparrow/keystoreimport/SwController.java
```diff
@@ -28,7 +28,9 @@ public void initializeView() {
             } else if(importer instanceof KeystoreXprvImport) {
                 importPane = new XprvKeystoreImportPane(getMasterController().getWallet(), (KeystoreXprvImport)importer, getMasterController().getDefaultDerivation());
             } else if(importer instanceof KeystoreMnemonicShareImport) {
-                importPane = new MnemonicShareKeystoreImportPane(getMasterController().getWallet(), (KeystoreMnemonicShareImport)importer, getMasterController().getDefaultDerivation());
+                importPane = new MnemonicShareKeystoreImportPane(getMasterController().getWallet(), (KeystoreMnemonicShareImport) importer, getMasterController().getDefaultDerivation());
```
craigraw: Nit: remove space after cast

src/main/java/com/sparrowwallet/sparrow/control/CodexKeystoreImportPane.java
```diff
@@ -0,0 +1,256 @@
...
+        enterCodexButton.managedProperty().bind(enterCodexButton.visibleProperty());
+        enterCodexButton.setOnAction(event -> {
+            enterCodex();
+        });
+
+        return enterCodexButton;
+    }
+
+    private void enterCodex() {
+        setDescription("Enter Codex32 secret share");
```
craigraw: Change to `"Enter secret share"`

src/main/java/com/sparrowwallet/sparrow/control/CodexKeystoreImportPane.java
```diff
@@ -0,0 +1,256 @@
...
+            boolean validChecksum = false;
+            onInputChange(empty, validChecksum);
+        });
+
+        HBox rightBox = new HBox();
+        rightBox.setSpacing(10);
+
+        calculateButton = new Button("Create Keystore");
+        calculateButton.setDisable(true);
+        calculateButton.setDefaultButton(!AppServices.onlineProperty().get());
```
craigraw: Change to:
```
calculateButton.setDefaultButton(true);
```

We always want this enabled.

src/main/java/com/sparrowwallet/sparrow/control/CodexKeystoreImportPane.java
```diff
@@ -0,0 +1,256 @@
...
+        buttonPane.getChildren().add(rightBox);
+        AnchorPane.setRightAnchor(rightBox, 0.0);
+
+        vBox.getChildren().add(buttonPane);
+
+        Platform.runLater(shareField::requestFocus);
+
+        StackPane stackPane = new StackPane();
+        stackPane.getChildren().add(vBox);
+        return stackPane;
```
craigraw: Replace with:
```
return vBox;
```

The `StackPane` is unnecessary and introduces unwanted layout changes.

ikmckenz:
Thank you for the review! I believe I have addressed all your comments.

src/main/java/com/sparrowwallet/sparrow/control/CodexKeystoreImportPane.java
```diff
@@ -0,0 +1,252 @@
...
+
+        calculateButton.setDisable(!validChecksum);
+        validLabel.setVisible(validChecksum);
+        invalidLabel.setVisible(!validChecksum && !empty);
+    }
+
+    private Node getSecretShareEntry() {
+        VBox vBox = new VBox(20);
+        vBox.setPadding(new Insets(10, 30, 10, 30));
+        vBox.setSpacing(10);
```
craigraw: Apologies, I missed this - this line should be removed. It's overriding the spacing set above in `new VBox(20)`.

craigraw:
Thanks @ikmckenz!

PR #1921, opened 2026-01-05
Add a RAW QR option to QRDisplayDialog

nymbus256:
This PR replaces the default toggle UR/BBQr behavior in the show QR dialog window with a drop down menu. The drop down menu offers UR, BBQr and RAW as options. 

RAW is useful when using sparrow as signer for watch only wallet apps (like Blue Wallet) that only support broadcasting transaction QRs with raw hex encoding.

image13025_clean

### DIFF
--- a/src/main/java/com/sparrowwallet/sparrow/AppController.java
+++ b/src/main/java/com/sparrowwallet/sparrow/AppController.java
@@ -771,7 +771,8 @@ public class AppController implements Initializable {
                 byte[] txBytes = transaction.bitcoinSerialize();
                 UR ur = UR.fromBytes(txBytes);
                 BBQR bbqr = new BBQR(BBQRType.TXN, txBytes);
-                QRDisplayDialog qrDisplayDialog = new QRDisplayDialog(ur, bbqr, false, false, false);
+                String txHex = Utils.bytesToHex(txBytes);
+                QRDisplayDialog qrDisplayDialog = new QRDisplayDialog(ur, bbqr, txHex, false, false, false);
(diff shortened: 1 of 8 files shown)


craigraw:
> RAW is useful when using sparrow as signer for watch only wallet apps (like Blue Wallet) that only support broadcasting transaction QRs with raw hex encoding.

This is only useful to use BW as a broadcast mechanism for signed transactions, correct? Have you explored the limits of what transaction size can be displayed in a single QR?

Wrt the drop-down, having the same graphic for each element really just adds to the visual complexity without helping with differentiation. Perhaps a better approach would be to use the logos from Blockchain Commons and Coinkite here. I also wonder if adding the word "format" to each item would help.

nymbus256:
> This is only useful to use BW as a broadcast mechanism for signed transactions, correct? 

Yes, this was the original intent.

> Have you explored the limits of what transaction size can be displayed in a single QR?

I have. This feature only works for small enough transactions and is grayed out (disabled) otherwise. Computing maximum transaction sizes depends on QR redundancy settings and constraints configured elsewhere. I avoid extra code and creating dependencies, by just checking if `getQrCode(rawTxData)` returns a valid QR or null.

> Wrt the drop-down, having the same graphic for each element really just adds to the visual complexity without helping with differentiation. Perhaps a better approach would be to use the logos from Blockchain Commons and Coinkite here. I also wonder if adding the word "format" to each item would help.

Thanks for the suggestions. I will look into them.

craigraw:
I've implemented this (and a few other related things) in 04de8370. Thanks for suggesting it.

PR #1932, opened 2026-01-17
user-agent for all exchange rate requests

lnliz:
this will add user-agent header for every http request in ExchangeSource file.

when i tested wallet export then the coingecko also complained about rate limits because of missing header.

should we add same http headers to feeSource file too ?

### DIFF
--- a/src/main/java/com/sparrowwallet/sparrow/net/ExchangeSource.java
+++ b/src/main/java/com/sparrowwallet/sparrow/net/ExchangeSource.java
@@ -68,7 +68,7 @@ public enum ExchangeSource {
 
             HttpClientService httpClientService = AppServices.getHttpClientService();
             try {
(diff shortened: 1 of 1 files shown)


craigraw:
Thanks, good catch - I missed the Coingecko historical rates. And the static variable is a good idea too.

> should we add same http headers to feeSource file too ?

I confirmed with the mempool.space team that they have no plans to require the User-Agent header. So it's not required in FeeRatesSource, and should be removed from the mempool.space calls here too.

lnliz:
ok, not sure why not add user agent to all http requests but ok with me, removed header from mempool.space calls

craigraw:
Thanks @lnliz. 

> not sure why not add user agent to all http requests

The main reason is privacy - even though the user agent is fairly generic it still provides some degree of potential fingerprinting.

PR #1944, opened 2026-02-10
work around import of corrupt samourai wallet backup file

ottosch:
There's a bug since latest Samourai Wallet releases that still exists in Ashigaru: the backup file is corrupted when saved and junk content is appended to it.

This PR works around it by, in case of JSONParseException, removing the content after the closing bracket.

I had to deal with this in a [ personal project](https://github.com/ottosch/brute-samourai/blob/master/src/samourai/samourai.go#L41) some time ago.


craigraw:
Thanks @ottosch!

PR #1946, opened 2026-02-11
.gitmodules: Sanitize.

apteryks:
This makes the git submodules work even if the top level project was cloned in a directory named something else than 'sparrowwallet'.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,6 +1,6 @@
 [submodule "drongo"]
 	path = drongo
-	url = ../../sparrowwallet/drongo.git
+	url = ../drongo.git
 [submodule "lark"]
 	path = lark
-	url = ../../sparrowwallet/lark.git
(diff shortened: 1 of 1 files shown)


craigraw:
The ../../sparrowwallet/ relative URL form is intentional. Relative submodule URLs resolve against the remote URL, not
   the local filesystem path, so the clone directory name doesn't matter. The current form ensures that forks of Sparrow
   still resolve the submodules to the sparrowwallet org rather than looking for drongo/lark under the forking user's
  account.

apteryks:
Hm. At least when using libgit2/guile-git I think this had caused problems; I've paged out the details but can try to reproduce the original issue I had.

apteryks:
I tried reproducing my original finding but couldn't; I'll close so far, and reopen if/when I have more data. Thanks!

Issue #1948, opened 2026-02-12
CA-based SSL certificate validation for Public Electrum servers


nzb-tuxxx:
In the current version, Sparrow does not strictly validate SSL certificates for public servers (self-signed certs can still be accepted). At the same time, more Electrum servers now provide valid CA-signed certificates, and certificate validation is generally best practice whenever technically possible.

 Could Sparrow provide an option (as shown in my screencast) that lets users choose to validate public server certificates against the system CA store (including hostname validation)?

https://github.com/user-attachments/assets/dd434935-51f6-43c6-bead-a1aba3c35682

  Possible default behavior:

  - Keep this option disabled by default (until all public servers provide valid certs), or
  - Enable/guide it based on a hardcoded allowlist of servers currently known to provide valid certs (e.g. Blockstream, DIYnodes).

  Benefits:

  - Improved privacy/security for users
  - Encourages Electrum server operators to deploy valid SSL certificates

  Note: I created an LLM-based PoC for this, but intentionally did not open a PR yet. I first wanted to check whether there is general interest in adding this option, and I assume you may have specific preferences for how this should be implemented.

craigraw:
I tried all the public servers to see which ones are still using self-signed certs, and it appears only electrum.bitaroo.net does. I've reached out to them to see if they would be interested in configuring a CA-verified TLS certificate.

That said, it's worth considering that if Sparrow starts requiring CA verified certs for public servers, existing installations will eventually fail to connect as the root certificates configured in `cacerts` within the bundled JVM expire. Generally, forcing the user to upgrade is avoided for Bitcoin software. I'm not sure how much of an issue it is here, but I think it's worth consideration.

nzb-tuxxx:
> I tried all the public servers to see which ones are still using self-signed certs, and it appears only electrum.bitaroo.net does. I've reached out to them to see if they would be interested in configuring a CA-verified TLS certificate.

I am unable to reproduce this. My result is completely different (only Blockstream and DIYNodes are fully valid), see below.

> That said, it's worth considering that if Sparrow starts requiring CA verified certs for public servers, existing installations will eventually fail to connect as the root certificates configured in cacerts within the bundled JVM expire. Generally, forcing the user to upgrade is avoided for Bitcoin software. I'm not sure how much of an issue it is here, but I think it's worth consideration.

[...]

PR #1952, opened 2026-02-17
Fix TLS certificate validation for Bitcoin RPC and public Electrum servers

kwsantiago:
- **BitcoindTransport**: Replace trust-all SSL factory with TOFU certificate pinning
- **TcpOverTlsTransport**: Use system CA trust managers for public Electrum servers


craigraw:
This should probably be two PRs, and there are several issues with the implementation on both items. Instead of merging this, I've implemented TOFU-based certificate pinning for TLS Bitcoin Core connections in 46d44461.

Wrt using system CA trust managers for public Electrum servers, this PR breaks connecting to the electrum.bitaroo.net server which is still using a self-signed cert. I'm tracking this issue in #1948.

PR #1953, opened 2026-02-17
Disable H2 verbose trace logging

kwsantiago:
## Summary
- Disable H2 database verbose trace logging (`TRACE_LEVEL_FILE=4` → `0`)
- Trace files (`.trace.db`) can contain sensitive wallet data and are readable even when the database is AES-encrypted
- Also removes the redundant `INIT=SET TRACE_LEVEL_FILE=4` from the JDBC URL


craigraw:
> Disable H2 database verbose trace logging (TRACE_LEVEL_FILE=4)

This is incorrect. `TRACE_LEVEL_FILE=4` is not "verbose trace logging". This setting routes logging output through SLF4J (the application's logging framework) and prevents .trace.db files from being created on disk. The current setting is actually the secure, intentional choice for a Bitcoin wallet application. What actually gets logged at each log level is managed through the Log4J configuration.

PR #1954, opened 2026-02-17
Bind embedded Electrum server to loopback only

kwsantiago:
## Summary
- Bind the embedded Electrum server socket to loopback (`127.0.0.1`) instead of all interfaces (`0.0.0.0`)
- Previously, other machines on the local network could connect and query wallet data


craigraw:
Good idea but there was a problem with the implementation - Sparrow connects to the IPv4 loopback while this might bind to the IPv6 loopback. Given this is a one line change I've made it in cd0be365.

PR #1957, opened 2026-02-23
Derive appropriate script type keystore from seed/xprv when importing Electrum wallet

busayo-OD:
This PR resolves TODO in Electrum.getKeystore() to derive the appropriate keystore for the requested script type when the wallet's script type or derivation path doesn't match.

Previously, any script type mismatch would throw an ImportException. Now, if the keystore has a seed or master private key, the correct keystore is derived using the standard BIP derivation path for the requested script type.

This also handles Electrum's non-standard derivation paths (e.g. m/0') by re-deriving to the standard path even when script types match.

Added tests:
- testSinglesigSeedKeystoreScriptTypeMismatch
- testSinglesigSeedKeystoreScriptTypeMatch

### DIFF
--- a/src/main/java/com/sparrowwallet/sparrow/io/Electrum.java
+++ b/src/main/java/com/sparrowwallet/sparrow/io/Electrum.java
@@ -47,12 +47,32 @@ public class Electrum implements KeystoreFileImport, WalletImport, WalletExport 
             throw new ImportException("Multisig wallet detected - import it using File > Import Wallet");
         }
 
-        if(!wallet.getScriptType().equals(scriptType)) {
-            //TODO: Derive appropriate ScriptType keystore from xprv if present
-            throw new ImportException("Wallet has an incompatible script type of " + wallet.getScriptType() + ", and the correct script type cannot be derived without the master private key.");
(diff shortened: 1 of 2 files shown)


craigraw:
Thanks for the PR. The core idea of deriving the correct keystore when the script type doesn't match has some merit, but I have concerns about the current implementation that prevent me from merging this as-is.

The main issue is the scope beyond what the TODO described. The TODO was specifically about handling script type mismatches, but this PR also silently re-derives the keystore when the script type matches but the derivation path is non-standard (e.g. Electrum's `m/0'`). If someone is deliberately using a non-standard derivation path in Electrum, Sparrow would silently override it to the standard path on import. This could cause the user to see different addresses than expected, which is a funds-loss risk. There should be no silent re-derivation — if anything, this situation warrants a warning to the user.

The tests also only verify the derivation path string rather than checking that the resulting keystore produces the expected xpub or addresses, which is what actually matters when re-deriving.

I'm going to close this for now, but if you'd like to revisit it I'd suggest limiting the scope to only re-deriving on an actual script type mismatch and adding tests that verify the resulting xpub.

PR #1959, opened 2026-02-25
Fix custom policy change detection in wallet settings

busayo-OD:
When editing a custom Miniscript policy in wallet settings, Sparrow should warn the user that their addresses will change. This was not happening because SettingsWalletForm.isAddressChange() had a TODO placeholder instead of an actual comparison.

This PR replaces the TODO with a Miniscript comparison, depending on equals() added in [drongo#40](https://github.com/sparrowwallet/drongo/pull/40).


craigraw:
Thanks, but `PolicyType.CUSTOM` is unused and will be removed - so it doesn't really make sense to add this.

PR #1960, opened 2026-03-01
show entire previous outpoint when possible

ottosch:
Hello, I would like to propose changing how previous outpoints are displayed: instead of being cut in the 8th char, they would take as much space as possible, displaying ellipsis (with CSS text-overflow) if there isn't enough room. The exception would be tooltips, which seem to always fit the whole prevout nicely.

I dealt with maximum width on OutputTableCell::formatCell too, although I haven't tested it in practice (I think sparrow server isn't used anymore).

I'm open to further adjustments :+1: .

image

image

image

image

image

image

### DIFF
--- a/src/main/java/com/sparrowwallet/sparrow/control/TransactionDiagram.java
+++ b/src/main/java/com/sparrowwallet/sparrow/control/TransactionDiagram.java
@@ -570,7 +570,7 @@ public class TransactionDiagram extends GridPane {
     }
 
     private String getInputDescription(BlockTransactionHashIndex input) {
-        return input.getLabel() != null && !input.getLabel().isEmpty() ? input.getLabel() : input.getHashAsString().substring(0, 8) + "..:" + input.getIndex();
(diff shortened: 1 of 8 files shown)


craigraw:
I don't think this change is an improvement unfortunately.

The current approach was specifically taken because the output index is _significantly_ more valuable that the 9th, 10th, 11th etc character in the transaction id hex. Note how the output index is cut off on the "TESTNET" screenshot above. With this proposal, you would have to give the Output column a much greater width (as seen on the following screenshot) to see the output index. 

I'm also not a fan of the appearance of the transaction diagram with this change. It's simply adding too much visual complexity which is distracting from the goal of the transaction diagram - to display clearly and simply the nature of the transaction.

ottosch:
No problem. Yeah, I tried to find a way to stretch/shrink in the middle, instead of at the end, but it seems there's no trivial way of doing it.

craigraw:
I agree, no simple way to do it. Closing this off.

PR #1962, opened 2026-03-05
fix: make password change atomic to prevent wallet DB corruption

### LINKED ISSUE #1961: Killing Sparrow immediately after removing wallet password can corrupt DB wallet (Error: Wallet was not encrypted, but seed is)
## Summary

There is a reproducible corruption bug in the Sparrow GUI password-change flow.

If the wallet password is removed in the GUI and Sparrow is killed immediately after confirming `No Password`, the wallet can end up in an inconsistent state:

  - the H2 wallet file is already unencrypted
  - but the seed row is still stored as encrypted

  After that, Sparrow fails to open the wallet with:

  `Wallet was not encrypted, but seed is`

  This makes the wallet file effectively unreadable by Sparrow.

  ## Impact

  - the wallet file can become unreadable by Sparrow
  - the user may be locked out of that wallet file
  - if the user still has the seed, funds are recoverable
  - if the corrupted wallet file was the only remaining copy of the seed material, this could become a serious loss-of-access scenario

  ## Environment

  - Sparrow Wallet GUI at eeade6a
  - software wallet with seed
  - reproduced on `testnet4`

  ## Reproduction Steps

  1. Create or open an encrypted software wallet with a seed in Sparrow.
  2. Open wallet settings.
  3. Start the password change / removal flow.
  4. Enter the current wallet password correctly.
  5. In the next dialog, leave the password empty and choose `No Password`.
  6. Immediately kill the Sparrow process right after clicking `No Password`.
     - Important: killing before the final click did **not** reproduce it.
     - The corruption happens when killing immediately **after** the final confirmation.
  7. Restart Sparrow and try to open the wallet.

  ## Actual Result

  Opening the wallet fails with:

  `Wallet was not encrypted, but seed is`

  ## Expected Result

  Removing the wallet password should be atomic and crash-safe.

  Either:

  - both file encryption and seed/keystore encryption are fully migrated, or
  - neither change is committed

  The wallet must not be left in a mixed state.

  ## Observed Corrupted State

  In the reproduced broken wallet:

  - the H2 `.mv.db` file is readable as an unencrypted DB
  - but in `wallet_master.seed`:
    - `mnemonicString = NULL`
    - `encryptedBytes` is still populated

  This matches the failure condition exactly.

  ## Suspected Root Cause

  From LLM code analysis, the password-removal flow appears to be non-atomic for DB wallets:

  - the DB/file encryption state is changed first
  - seed/keystore encryption updates are persisted later through an asynchronous background update path

  That creates a crash window where:

  - the wallet file is already unencrypted
  - but the seed is still stored as encrypted

  Relevant code paths:

  - `src/main/java/com/sparrowwallet/sparrow/wallet/SettingsController.java`
  - `src/main/java/com/sparrowwallet/sparrow/wallet/SettingsWalletForm.java`
  - `src/main/java/com/sparrowwallet/sparrow/io/db/DbPersistence.java`
  - `src/main/java/com/sparrowwallet/sparrow/io/db/KeystoreDao.java`
  - `src/main/java/com/sparrowwallet/sparrow/io/Storage.java`

  ## Why this matters

  A user can trigger this during a normal GUI workflow. If Sparrow crashes, is force-killed, or the system shuts down at the wrong time, the wallet file can become unusable.

  ## Suggested Fix Direction

  - make password removal / password change atomic for DB wallets
  - do not rely on async background persistence for the critical seed/keystore migration
  - ensure file encryption change and seed/keystore encryption change are committed together
  - add a regression test for crash/interruption during password removal

  ## Additional Material

  I can provide a throwaway corrupted `testnet4.mv.db` wallet file that reproduces the issue for debugging.

  It contains no coins and was created only for reproducing this bug.

PrinceOfEgypt:
## Summary

Fixes #1961

When removing or changing a wallet password, the previous code called `updatePassword()` (which re-encrypts the H2 `.mv.db` file on disk) **synchronously on the calling thread**, then queued seed/keystore row updates on a background executor. If Sparrow was killed in the window between those two steps, the wallet ended up in a mixed state:

- the DB file was already unencrypted (or re-encrypted with the new key)
- but the seed row in `wallet_master.seed` still held `encryptedBytes` from the old password

On next open this produces: `Wallet was not encrypted, but seed is`

## Root cause

`DbPersistence.updateWallet()` — the file encryption change and the seed/keystore DB row update were **not atomic**:

```java
// BEFORE (broken)
updatePassword(storage, encryptionPubKey);   // synchronous — H2 file re-encrypted NOW
updateExecutor.execute(() -> {               // async — seed rows updated LATER (crash window)
    update(storage, wallet, ...);
});
```

Additionally, `keystoreEncryptionChanged()` itself queues the dirty keystores onto the same executor, so the seed rows weren't even guaranteed to be in the dirty map yet when `updatePassword()` ran.

## Fix

When the file password is actually changing, defer `updatePassword()` onto the executor **after** a new `flushEncryptionKeystores()` step. Because the executor is single-threaded and FIFO, this guarantees ordering:

1. `keystoreEncryptionChanged` task → `encryptionKeystores` populated in dirty map
2. `flushEncryptionKeystores()` → seed/keystore rows written with **old** file password
3. `updatePassword()` → H2 file encryption changed
4. `update()` → remaining dirty fields flushed with new password

There is now no crash window where file and row state can diverge. The no-password-change path is unchanged.

## Testing

1. Create an encrypted software wallet with a seed
2. Open wallet settings → change to no password → immediately kill Sparrow after confirming
3. Restart → wallet should open cleanly with no password required

### DIFF
--- a/src/main/java/com/sparrowwallet/sparrow/io/db/DbPersistence.java
+++ b/src/main/java/com/sparrowwallet/sparrow/io/db/DbPersistence.java
@@ -162,13 +162,65 @@ public class DbPersistence implements Persistence {
 
     @Override
     public void updateWallet(Storage storage, Wallet wallet, ECKey encryptionPubKey) throws StorageException {
-        updatePassword(storage, encryptionPubKey);
+        String newPassword = getFilePassword(encryptionPubKey);
 
-        updateExecutor.execute(() -> {
+        if(dataSource != null && !Objects.equals(getDatasourcePassword(), newPassword)) {
+            // Password is changing: capture the current password now, then run the keystore-encryption
+            // flush and the H2 file re-encryption atomically on the executor.  This guarantees that any
+            // pending keystoreEncryptionChanged task (which is also queued on the executor) completes
+            // first, so seed/keystore rows are written under the OLD file password before the file
+            // encryption is changed – eliminating the crash-corruption window where the DB file is
+            // already unencrypted but the seed row still holds encrypted data.
+            String currentPassword = getDatasourcePassword();
+            updateExecutor.execute(() -> {
+                try {
+                    flushEncryptionKeystores(storage, wallet, currentPassword);
+                    updatePassword(storage, encryptionPubKey);
+                    update(storage, wallet, newPassword);
+                } catch(Exception e) {
+                    log.error("Error updating wallet db during password change", e);
+                }
+            });
+        } else {
+            updatePassword(storage, encryptionPubKey);
+            updateExecutor.execute(() -> {
+                try {
+                    update(storage, wallet, newPassword);
+                } catch(Exception e) {
+                    log.error("Error updating wallet db", e);
+                }
+            });
+        }
+    }
+
+    /**
+     * Synchronously writes any pending keystore/seed encryption changes to the DB using the supplied
+     * (current/old) file password, then clears them from the dirty map so {@link #update} does not
+     * process them a second time.  Must be called from the update executor thread before the file
+     * encryption is changed.
+     */
+    private void flushEncryptionKeystores(Storage storage, Wallet wallet, String currentPassword) throws StorageException {
+        DirtyPersistables dirtyPersistables = dirtyPersistablesMap.get(wallet);
+        if(dirtyPersistables == null || dirtyPersistables.encryptionKeystores.isEmpty()) {
+            return;
+        }
+
+        List<Keystore> keystoresToFlush = new ArrayList<>(dirtyPersistables.encryptionKeystores);
+        dirtyPersistables.encryptionKeystores.clear();
+
+        log.debug("Flushing " + keystoresToFlush.size() + " keystore encryption change(s) for " + wallet.getFullName() + " before file password change");
+
+        Jdbi jdbi = getJdbi(storage, currentPassword);
+        jdbi.useHandle(handle -> {
+            WalletDao walletDao = handle.attach(WalletDao.class);
             try {
-                update(storage, wallet, getFilePassword(encryptionPubKey));
-            } catch(Exception e) {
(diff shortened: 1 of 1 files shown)


craigraw:
Thanks for the report and analysis. The underlying issue — `updatePassword()` running before the seed/keystore rows are updated — is real, and I've addressed it in e9108b85.

The approach I've taken is simpler: instead of adding a new `flushEncryptionKeystores()` method, I moved both `updatePassword()` and `update()` onto the executor and made the ordering direction-dependent. When removing encryption, data is written first (using the old password), then the file encryption is removed. When adding encryption, the file is encrypted first, then data is written. Both orderings ensure that a crash between the two steps leaves the wallet in a recoverable state rather than an unreadable one.

A couple of notes on the PR as submitted:

- The ordering here is always data-first (flush keystores → change file → update remaining). This is safe for password removal, but if the same path is hit during password addition (encrypted seed written to an unencrypted file, then crash before file encryption), the result is the same unreadable state the fix is trying to prevent. The direction-dependent ordering avoids this.
- Calling it "atomic" is a bit of a misnomer — the operations are still sequential and a crash between them still leaves an inconsistent (though recoverable) state. True atomicity would require something like writing to a temp file and doing an atomic rename.

PrinceOfEgypt:
Thanks for the quick turnaround on this, and for the thorough feedback.

The direction-dependent ordering is a cleaner solution — I hadn't considered the symmetric failure case during password *addition*, so that's a good catch. And the point about "atomic" being a misnomer is well taken; "crash-safe" would have been more accurate given that the operations are still sequential.

Glad the underlying analysis was useful. If you're interested in pursuing the temp-file + atomic rename approach as a future hardening step, happy to put together another PR.

Issue #1896, opened 2025-12-04
Add optional HMAC challenge-response as 2FA for wallet encryption


kwsantiago:
Support combining password + YubiKey HMAC-SHA1 challenge-response for wallet encryption key derivation.

- Optional second factor alongside existing password
- Uses YubiKey's offline HMAC challenge-response (slot 2)
- Final key = derive(password) + HMAC(challenge)

Similar to KeePassXC and LUKS yubikey-full-disk-encryption implementations.

References:
- https://docs.yubico.com/yesdk/users-manual/application-otp/challenge-response.html
- https://github.com/keepassxreboot/keepassxc (KeePassXC implementation)
- https://github.com/agherzan/yubikey-full-disk-encryption (LUKS implementation)

kwsantiago:
@craigraw thoughts? I'd like to see this implemented, happy to work on a PR if it's something you'd consider adding into Sparrow.

craigraw:
Note: this is related to #1561.

I'd like to understand what UI flow is necessary to set this up and, once setup, manage password entry.

kwsantiago:
@craigraw what I envision is an optional 2FA feature that prompts for a YubiKey (or in a separate issue 2FA generated code) after the password is entered, acting as an additional layer of security for the wallet.

craigraw:
Understood. But what does the setup for this look like - it is just another checkbox on the password confirmation dialog? And would this be Yubikey specific, or would the challenge-response framework be generalizable to other devices?

kwsantiago:
@craigraw For setup: Checkbox on password dialog, then touch YubiKey to register.

For unlock: Password first, then touch YubiKey if configured. Challenge = wallet salt.

I'm thinking we should have a simple, generalized interface that other HMAC devices could implement later, and use YubiKey as a starting point.

kirillkovalenko:
I suggest considering a more general 'multiple key protectors' architecture used by LUKS, BitLocker and others, where the data encryption key is wrapped by one or more independent protector-derived keys.

Protectors can be, but not limited to:

- Null protector
- Password protector
- YubiKey HMAC-SHA1 challenge-response protector
- FIDO2 'hmac-secret' protector
- Smart-card / PIV
- PGP card
- Key file
- Flash drive/SD card (HW fingerprint + optional key file)

[...]

PR #1967, opened 2026-03-11
Add optional challenge-response for wallet encryption

kwsantiago:
## Summary

Adds an optional challenge-response second factor for wallet encryption, addressing #1896.

When enabled the wallet encryption key becomes `SHA256(argon2DerivedKey || HMAC_SHA1(argon2Salt))`. Both password plus challenge-response and challenge-response only (no password) are supported. A new `SPRW2` wallet header records the requirement so it is known before the password prompt.

Backed by YubiKey slot 2 and any device speaking the same protocol. The UI is vendor neutral.

## Dependencies

- sparrowwallet/drongo#41 — key derivation interfaces
- sparrowwallet/lark#4 — HMAC-SHA1 USB implementation

The `drongo` and `lark` submodule pointers are deliberately not bumped here, as they would point at fork commits. CI will not compile until those land and the pointers move.

## Design

**Header.** `SPRW2` adds a flags byte between the magic and the salt, bit 0 meaning challenge-response is required. Unrecognized flag bits are rejected rather than ignored. An older Sparrow reads an `SPRW2` file as an invalid password rather than opening it.

**Provider wiring.** The provider comes from a factory registered once at desktop startup and is obtained inside `DbPersistence` when the header says it is required, so no unlock path can skip the second factor by omitting an injection. Terminal mode registers no factory and reports the requirement as unsupported.

**Failure handling.** A device error surfaces with its own message rather than as an unsupported file format. A wrong password is still reported as a wrong password. Dismissing the touch prompt interrupts the wait. A failed derive while enabling or disabling restores the in-memory state to match the file.

## Security properties

- The challenge is the wallet's Argon2 salt, stored in the header and never rotated, so the response is a static secret for that wallet. If slot 2 requires no button press, anything with USB access can obtain it. Inherent to HMAC-SHA1 challenge-response.
- Enabling binds the file to that device. There is no second-device registration or recovery export, so losing or reprogramming the key makes the file unopenable. The dialog warns and asks for confirmation before enabling or removing.
- The flags byte is unauthenticated. Altering it derives a key that fails to open the file: denial of service, not bypass.

## Tests

`DbPersistenceHeaderTest` covers the header parse: `SPRW1` and `SPRW2` both leaving the stream at the correct salt offset, set and cleared flags, rejection of unrecognized bits, and rejection of a file truncated after the magic.

`DbPersistenceChallengeResponseTest` covers the derivation decisions against a stub provider: that an empty password on a challenge-response wallet does not return the no-password key, that it still does on an ordinary wallet, that derivation fails rather than proceeding without a device when no provider is registered, that the response changes the derived key, and that a fresh provider is obtained per derivation so a retry after a failed touch is not silently downgraded.

Composite build against current master of all three repos compiles and passes, 862 tests.

Manually exercised with a YubiKey: password-only wallet, enabling challenge-response, opening with password plus touch, challenge-response only, changing the password on an enabled wallet, device absent, and touch timeout.

Challenge padding follows `ykman`, so a slot programmed for variable-length challenges resolves the challenge boundary correctly. Details on sparrowwallet/lark#4.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -134,7 +134,7 @@ processResources {
 
 test {
     useJUnitPlatform()
-    jvmArgs = ["--add-opens=java.base/java.io=ALL-UNNAMED", "--enable-native-access=ALL-UNNAMED"]
+    jvmArgs = ["--add-opens=java.base/java.io=ALL-UNNAMED", "--add-opens=java.base/java.math=ALL-UNNAMED", "--enable-native-access=ALL-UNNAMED"]
 }
 
 application {
@@ -162,6 +162,7 @@ application {
                                  "--add-opens=javafx.graphics/javafx.scene.input=com.sparrowwallet.sparrow",
                                  "--add-opens=java.base/java.net=com.sparrowwallet.sparrow",
                                  "--add-opens=java.base/java.io=com.google.gson",
+                                 "--add-opens=java.base/java.math=com.sparrowwallet.drongo",
                                  "--add-opens=java.smartcardio/sun.security.smartcardio=com.sparrowwallet.sparrow",
                                  "--add-reads=kotlin.stdlib=kotlinx.coroutines.core",
                                  "--add-reads=org.flywaydb.core=java.desktop"]
@@ -212,6 +213,7 @@ jlink {
                    "--add-opens=javafx.graphics/com.sun.javafx.application=com.sparrowwallet.sparrow",
                    "--add-opens=java.base/java.net=com.sparrowwallet.sparrow",
                    "--add-opens=java.base/java.io=com.google.gson",
+                   "--add-opens=java.base/java.math=com.sparrowwallet.drongo",
                    "--add-opens=java.smartcardio/sun.security.smartcardio=com.sparrowwallet.sparrow",
                    "--add-reads=com.sparrowwallet.merged.module=java.desktop",
                    "--add-reads=com.sparrowwallet.merged.module=java.sql",
--- a/src/main/java/com/sparrowwallet/sparrow/AppController.java
+++ b/src/main/java/com/sparrowwallet/sparrow/AppController.java
@@ -1205,7 +1205,7 @@ public class AppController implements Initializable {
 
                 SecureString password = optionalPassword.get();
                 if(storage.isChallengeResponseEnabled()) {
-                    storage.setChallengeResponseProvider(com.sparrowwallet.sparrow.AppServices.createYubiKeyProvider());
+                    storage.setChallengeResponseProvider(AppServices.createYubiKeyProvider());
                 }
                 Storage.LoadWalletService loadWalletService = new Storage.LoadWalletService(storage, password);
                 loadWalletService.setOnSucceeded(workerStateEvent -> {
@@ -1404,7 +1404,7 @@ public class AppController implements Initializable {
             } else {
                 if(dlg.isYubikeyEnabled()) {
                     storage.setChallengeResponseEnabled(true);
-                    storage.setChallengeResponseProvider(com.sparrowwallet.sparrow.AppServices.createYubiKeyProvider());
+                    storage.setChallengeResponseProvider(AppServices.createYubiKeyProvider());
                 }
                 keyDerivationService = new Storage.KeyDerivationService(storage, password.get());
                 keyDerivationService.setOnSucceeded(workerStateEvent -> {
@@ -2346,7 +2346,7 @@ public class AppController implements Initializable {
                 Optional<SecureString> password = dlg.showAndWait();
                 if(password.isPresent()) {
                     if(storage.isChallengeResponseEnabled()) {
-                        storage.setChallengeResponseProvider(com.sparrowwallet.sparrow.AppServices.createYubiKeyProvider());
+                        storage.setChallengeResponseProvider(AppServices.createYubiKeyProvider());
                     }
                     keyDerivationService = new Storage.KeyDerivationService(storage, password.get(), true);
                     keyDerivationService.setOnSucceeded(workerStateEvent -> {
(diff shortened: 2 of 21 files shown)


nzb-tuxxx:
I build a custom version `sparrow-wallet-yubikey-hmac-challenge-response-git-0.r2286.gdfdca95-1-x86_64.pkg.tar.zst` for arch based on this branch for testing, but it crashes at the point where checkbox is set for yubikey in the password dialog:

```
2026-03-11 19:23:11,975 ERROR Exception in thread "JavaFX Application Thread"
java.lang.reflect.InaccessibleObjectException: Unable to make field final int[] java.math.BigInteger.mag accessible: module java.base does not "opens java.math" to module com.sparrowwallet.drongo
```
Probably a missing dependency within Gradle-Setup? Could you fix that or did I do something wrong on my side?

Also I noticed that the password set dialog changes unintentionally between "Set yubikey" and "set password" if you first checkbox "Require yubikey for unlock" and then enter a password afterwards.

kwsantiago:
@nzb-tuxxx Thank you for testing and reporting! The crash was a missing `--add-opens` in Sparrow's Gradle config, since drongo needs reflective access to `BigInteger` for zeroing key material. Nothing wrong on your side. The button text bug is fixed too. Both are on the branch if you want to rebuild.

nzb-tuxxx:
Thanks, I got it working and successfully managed a YubiKey encrypt / decrypt for a testing wallet. What I noticed:

1. Change password does not work anymore. It only asks for current password but does not accept it (`The password was incorrect`). IMO there should be 3 ways for changing: No password, YubiKey only, YubiKey + new password. What about new YubiKey: probably unsupported?  
2. If YubiKey is not connected while unlock, error message is `com.sparrowwallet.drongo.crypto.KeyCrypterException: Challenge-response key derivation failed` - if it is possible to detect that there is no yubikey connected, can we improve the message accordingly ("No YubiKey found, plug-in YubiKey and try again")? 
3. If YubiKey is not pressed within a time frame of 15 seconds, some kind of timeout occurs. Is this intentionally? The YubiKey stops blinking and does not recognize touch afterwards, while the "Touch your YubiKey" Dialog is still displayed. On console I see `Caused by: com.sparrowwallet.drongo.crypto.ChallengeResponseException: Failed to claim YubiKey interface: LIBUSB_ERROR_BUSY` - maybe this is due to a broken USB Virtualization in VirtualBox. I need to restart Sparrow then in order to try again, all retries without sparrow restart fail

kwsantiago:
@nzb-tuxxx Thank you for the thorough testing! All three are fixed:

1. Change password: the verify step was accidentally disabling the YubiKey provider before key derivation, so it always mismatched. Fixed.
2. Error messages: the specific YubiKey errors (e.g. "No YubiKey found") were getting swallowed by a generic wrapper. Now they pass through to the dialog.
3. Timeout/LIBUSB_ERROR_BUSY: the pending YubiKey operation wasn't being cancelled on timeout, leaving the USB interface locked. Fixed, though VirtualBox USB passthrough may still add its own quirks.

Please let me know if you still see these or other issues.

nzb-tuxxx:
Thanks, 1 and 2 looking good. But 3 is still unchanged. Need to restart sparrow to be able to use yubikey again after 15 seconds timeout. I'll test this on real hardware, just to be sure.

---

Edit: Also happens on real hardware

craigraw:
Thanks for this. I think this would be a good addition to Sparrow in principle, but there are significant code changes which need careful review - I haven't attempted this yet.

There is however a problem which needs addressing. Currently, there is a UI change to add a checkbox with the text `Require YubiKey for unlock`. This ties the Sparrow UI to a specific vendor, which I dislike on principle. But further, how are other vendors then accommodated? For example, both Trezor (FIDO2) and Ledger (U2F) support challenge/response authentication, and are probably going to end up being used more than the Yubikey. Until this is addressed I think the PRs should be marked as draft.

A further question: what value is `FLAG_CHALLENGE_RESPONSE` adding?

nzb-tuxxx:
> This ties the Sparrow UI to a specific vendor, which I dislike on principle

Agree, besides the mentioned hardware wallets, there is also [OnlyKey](https://onlykey.io/) which supports the same protocol: https://docs.onlykey.io/usersguide.html#using-onlykey-with-a-software-password-manager - Maybe just name it `2FA (Challenge-Response)` with a question-mark that links to sparrow wallet faq for an entry that should be created in the future, before merging, that explains the whole process - maybe https://sparrowwallet.com/docs/faq.html#how-do-i-use-2fa-for-wallet-encryption)?

kwsantiago:
@craigraw Agreed and addressed, all user-facing strings now say "challenge-response" / "security key" instead of "YubiKey". `FLAG_CHALLENGE_RESPONSE` is needed in the header so Sparrow knows at load time to set up the provider and allow empty passwords, otherwise key derivation silently produces the wrong key.

@nzb-tuxxx on issue 3, this should be fixed now. The poll loop wasn't detecting the YubiKey's internal ~15s timeout, so it held the USB interface locked. Now detects it immediately and cleans up. Should be retryable without restart. Can you try it again?

Agreed regarding OnlyKey, they use the same HMAC-SHA1 protocol, and the `ChallengeResponseProvider` interface makes it straightforward to add support for other devices.

nzb-tuxxx:
With [5accef3](https://github.com/sparrowwallet/sparrow/pull/1967/commits/5accef3992d6cc47101fbee7ba7882f930769155) every touch responses with `Security key timed out..` . Not sure if i memorized my test password wrong or if this build is broken

kwsantiago:
@nzb-tuxxx Bug in the timeout detection. After a successful touch, `RESP_TIMEOUT_WAIT_FLAG` clears at the same time `RESP_PENDING_FLAG` sets, but the code was checking for timeout before checking for a response. Fixed in de070b1, should work now.

nzb-tuxxx:
Timeout is working now as expected, thanks for your effort @kwsantiago

kirillkovalenko:
@craigraw what is the status of this pr? are you going to merge it any time soon?

PR #2008, opened 2026-05-26
show full year on balance chart x-axis

### LINKED ISSUE #1703: show year on balance plot horizontal axis
In a variety of places in sparrow wallet (i.e. Transactions tab and Wallet Summary screen), we have plots with a balance history vs time.

We do not have the year in the horizontal axis on those plots. Would like to see the year shown as well.

PeterXMR:
Fixes #1703.

`DateAxisFormatter.MONTH_FORMAT` was `MMM yy`; switched to `MMM yyyy` so multi-year balance charts (Transactions tab + Wallet Summary) show an unambiguous 4-digit year. Sub-year (`d MMM`) and sub-day (`HH:mm`) buckets are unchanged. Four locale-stable unit tests added.


craigraw:
Thanks, good improvement on the current chart.

PR #2014, opened 2026-05-28
Validate auth47 callback protocols

Asylian21:
## Summary
- reject non-HTTP/S auth47 callbacks when parsing the URI
- keep Soroban callbacks unchanged and use protocol-aware checks for normal callbacks
- add a regression test for callbacks that include a resource parameter

## Why
Auth47 responses are sent through `HttpURLConnection`, but a non-HTTP callback could be accepted when the URI supplied an `r` parameter. Validating the callback protocol during parsing keeps invalid callbacks from reaching the send path.

## Validation
- `git diff --check`
- `JAVA_HOME=/opt/homebrew/opt/openjdk@25/libexec/openjdk.jdk/Contents/Home ./gradlew :test --tests com.sparrowwallet.sparrow.net.Auth47Test --no-daemon --rerun-tasks`


craigraw:
Closed, see https://github.com/sparrowwallet/sparrow/pull/2016#issuecomment-4582436277

PR #2015, opened 2026-05-28
Validate LNURL-auth URL schemes

Asylian21:
## Summary
- reject LNURL-auth values that are not encoded with the `lnurl` Bech32 prefix
- enforce the LNURL transport boundary before any signing flow starts: HTTPS URLs are accepted, and HTTP is accepted only for onion hosts
- add focused regression coverage for HTTPS, HTTP onion, HTTP clearnet, non-HTTP, and non-LNURL Bech32 inputs

## Why
LUD-01 defines LNURL as a Bech32-encoded HTTPS or onion URL, and LUD-04 has wallets sign `k1` and send the signature/key back to the decoded LNURL endpoint. The existing parser accepted other schemes such as clearnet HTTP or FTP once `tag=login` and a valid `k1` were present. Rejecting these during parsing keeps unsafe callbacks out of the authentication/signing path.

References:
- https://github.com/lnurl/luds/blob/luds/01.md
- https://github.com/lnurl/luds/blob/luds/04.md

## Validation
- Before the fix, `JAVA_HOME=/opt/homebrew/opt/openjdk@25/libexec/openjdk.jdk/Contents/Home ./gradlew :test --tests com.sparrowwallet.sparrow.net.LnurlAuthTest --no-daemon --rerun-tasks` failed on `rejectsHttpClearnetCallbacks` and `rejectsNonHttpCallbacks`
- After the fix, the same command passes
- `git diff --check`
- `JAVA_HOME=/opt/homebrew/opt/openjdk@25/libexec/openjdk.jdk/Contents/Home ./gradlew :compileJava :compileTestJava --no-daemon`

### DIFF
--- a/src/main/java/com/sparrowwallet/sparrow/net/LnurlAuth.java
+++ b/src/main/java/com/sparrowwallet/sparrow/net/LnurlAuth.java
@@ -41,9 +41,17 @@ public class LnurlAuth {
     public LnurlAuth(URI uri) throws MalformedURLException, URISyntaxException {
         String lnurl = uri.getSchemeSpecificPart();
         Bech32.Bech32Data bech32 = Bech32.decode(lnurl, 2000);
+        if(!"lnurl".equals(bech32.hrp)) {
+            throw new IllegalArgumentException("LNURL-auth bech32 prefix must be lnurl.");
+        }
+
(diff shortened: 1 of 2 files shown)


craigraw:
Closed, see https://github.com/sparrowwallet/sparrow/pull/2016#issuecomment-4582436277.

PR #2016, opened 2026-05-28
Validate Payjoin URL schemes

Asylian21:
## Summary
- Parse the raw BIP21 pj parameter and accept only HTTPS endpoints, or HTTP endpoints for .onion hosts.
- Reuse the same validation when caching scanned Payjoin URIs and when sending the Payjoin PSBT request.
- Add focused regression coverage for valid HTTPS, valid HTTP .onion, non-HTTP .onion rejection, and malformed pj rejection.

## Why
BIP78 defines pj as an http(s) endpoint and requires senders to avoid unauthenticated or unencrypted connections unless the endpoint is protected by an authenticated mechanism such as a hidden service. Sparrow's previous validation came from BitcoinURI#getPayjoinUrl(), which allowed any scheme when the host ended in .onion and could throw on malformed or non-hierarchical pj values.

Reference: https://github.com/bitcoin/bips/blob/master/bip-0078.mediawiki

## Validation
- Before the fix, JAVA_HOME=/opt/homebrew/opt/openjdk@25/libexec/openjdk.jdk/Contents/Home ./gradlew :test --tests com.sparrowwallet.sparrow.AppServicesTest --no-daemon --rerun-tasks failed on the non-HTTP .onion and malformed pj cases.
- After the fix, the same targeted test command passes.
- git diff --check passes.

### DIFF
--- a/src/main/java/com/sparrowwallet/sparrow/AppServices.java
+++ b/src/main/java/com/sparrowwallet/sparrow/AppServices.java
@@ -842,8 +842,36 @@ public class AppServices {
         return payjoinURIs.get(address);
     }
 
+    public static URI getSecurePayjoinUrl(BitcoinURI bitcoinURI) {
+        if(bitcoinURI != null) {
+            Object payjoinUrlObj = bitcoinURI.getParameterByName(BitcoinURI.FIELD_PAYJOIN_URL);
+            if(payjoinUrlObj instanceof String payjoinUrl) {
+                try {
+                    URI uri = new URI(payjoinUrl);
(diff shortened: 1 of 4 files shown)


craigraw:
This PR, and the related #2014 and #2015 all do much the same thing, but create new near-identical helpers. I've refactored this in a single helper in https://github.com/sparrowwallet/drongo/commit/7d0699faa0d89a04c37425d018f143a4e26b49b9 and updated the call sites in Sparrow in 464fade6, implementing the suggestions here in a reusable way.

Thanks for the idea though.

PR #2017, opened 2026-05-28
Bound BBQr decoded payload size

Asylian21:
## Summary
- cap decoded BBQr payloads at 500,000 bytes before parsing PSBT, transaction, text, or binary data
- replace zlib readAllBytes() inflation with bounded streaming inflation
- guard multi-part concatenation before allocating the aggregate decoded buffer
- add regression coverage for oversized zlib and base32 BBQr payloads

## Why
BBQr input is accepted from scanned QR data. The BBQr spec targets PSBT and signed transaction payloads up to 500k, and mode Z decompresses the whole file after all QR parts are received. Without a decoded-output cap, an attacker-controlled BBQr payload can drive excessive allocation during decompression or assembly before normal parsing rejects it.

Spec reference: https://bbqr.org/BBQr.html

## Validation
- Before the fix, a focused oversized-zlib regression test failed because inflate() returned data above the intended cap.
- `JAVA_HOME=/opt/homebrew/opt/openjdk@25/libexec/openjdk.jdk/Contents/Home ./gradlew :test --tests com.sparrowwallet.sparrow.io.bbqr.BBQREncodingTest --tests com.sparrowwallet.sparrow.io.bbqr.BBQRDecoderTest --no-daemon --rerun-tasks`
- `git diff --check`

### DIFF
--- a/src/main/java/com/sparrowwallet/sparrow/io/bbqr/BBQRDecoder.java
+++ b/src/main/java/com/sparrowwallet/sparrow/io/bbqr/BBQRDecoder.java
@@ -37,9 +37,7 @@ public class BBQRDecoder {
 
             if(receivedParts.size() == totalParts) {
                 byte[] data = concatParts();
-                BBQREncoding.checkDecodedLength(data.length);
                 data = header.inflate(data);
-                BBQREncoding.checkDecodedLength(data.length);
 
                 if(type == BBQRType.PSBT) {
                     result = new Result(new PSBT(data));
--- a/src/main/java/com/sparrowwallet/sparrow/io/bbqr/BBQRDecoder.java
+++ b/src/main/java/com/sparrowwallet/sparrow/io/bbqr/BBQRDecoder.java
@@ -37,7 +37,9 @@ public class BBQRDecoder {
 
             if(receivedParts.size() == totalParts) {
                 byte[] data = concatParts();
+                BBQREncoding.checkDecodedLength(data.length);
                 data = header.inflate(data);
+                BBQREncoding.checkDecodedLength(data.length);
 
                 if(type == BBQRType.PSBT) {
                     result = new Result(new PSBT(data));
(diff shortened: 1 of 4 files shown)


craigraw:
Thanks for the patch.

To be upfront about the threat model: I don't think there's a live DoS vector here in practice. BBQr input comes from the camera, so an attacker first has to get the user to scan an attacker-controlled animated QR sequence. A single QR caps the compressed payload at ~3 KB, which inflates to a few MB at best, which is noise on a desktop JVM. The multi-part ceiling (`seqTotal` is base36, max 1295 parts) is the only path to a meaningfully large inflate, and it requires the user to patiently scan over a thousand frames of a sequence they were already tricked into trusting. Worst case is an app crash, not key exfiltration or wallet corruption.

That said, happy to look at this as hardening. A few small things:

1. Drop the per-part `checkDecodedLength` call added to `BASE32`'s `decode` override. A single QR is bounded by the QR spec well below the cap, so the per-part check can never fire — `concatParts` is what actually enforces the invariant. (HEX was left alone for the same reason, so this just brings BASE32 back into line.)
2. The two `BBQREncoding.checkDecodedLength(data.length)` calls in `BBQRDecoder.receivePart` are redundant — `concatParts` enforces during accumulation, and both the streaming `ZLIB.inflate` and the base-class `inflate` enforce on their own outputs. Pick one layer to own the invariant rather than three.
3. The line spacing in `rejectsOversizedZlibInflate` is odd.

Asylian21:
Thanks for the thorough review and the threat-model context — agreed this is hardening rather than a live DoS, with the multi-part inflate ceiling being the only meaningful path.

All three points are addressed in 8a475ae:

1. Dropped the per-part `checkDecodedLength` from the `decode()` overrides. One small correction: `HEX` had actually picked up the same per-part check, so I removed it from both `HEX` and `BASE32` to keep them in line. A single QR is bounded well below the cap, so `concatParts` is the sole enforcer on the raw concatenated data.

2. Removed both `checkDecodedLength` calls in `BBQRDecoder.receivePart`. The invariant is now owned by two layers only: `concatParts` bounds the pre-inflate concatenated buffer, and the inflate implementations bound their own output (the streaming `ZLIB.inflate`, and the base-class identity `inflate`). `receivePart` no longer re-checks.

3. Tidied the spacing in `rejectsOversizedZlibInflate`.

`BBQREncodingTest` and `BBQRDecoderTest` still pass — the oversized-zlib case is now caught directly by the streaming inflate.

craigraw:
I've actually reconsidered this. The 500kB mentioned in the spec is informal text, not a hard limit. Many PSBTs are over this size, and this change would prevent them from being ingested. As such I think this change does more harm than good. Apologies for not considering this in my first reply.

The streaming inflate change itself is an improvement over readAllBytes(), and I may address that separately in a later update. Appreciate the effort regardless.

PR #2019, opened 2026-06-01
Replace openpnp-capture-java dependency with io.github.doblon8 implementation

doblon8:
This PR replaces the `org.openpnp:openpnp-capture-java` dependency with `io.github.doblon8:openpnp-capture-java` (https://github.com/doblon8/openpnp-capture-java).

The new implementation uses Java's Foreign Function & Memory (FFM) API instead of JNA while providing equivalent functionality.

The native binaries bundled in `io.github.doblon8:openpnp-capture-java` were taken from https://github.com/sparrowwallet/openpnp-capture.

Tested on Linux (Debian 13) x86_64 and Windows 10 x86_64.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -205,7 +205,7 @@ jlink {
                    'glob:/org.hid4java/darwin-*/**,' +
                    'glob:/org.hid4java/linux-*/**,' +
                    'glob:/org.hid4java/win32-*/**,' +
-                   'glob:/io.github.doblon8.jzbar/native/**' +
+                   'glob:/io.github.doblon8.jzbar/native/**,' +
                    'glob:/io.github.doblon8.openpnp.capture/native/**']
     launcher {
         name = 'sparrow'
--- a/build.gradle
+++ b/build.gradle
@@ -73,7 +73,7 @@ dependencies {
     implementation('com.fasterxml.jackson.core:jackson-databind:2.21.1')
     implementation('com.sparrowwallet:hummingbird:1.7.4')
     implementation('co.nstant.in:cbor:0.9')
-    implementation('org.openpnp:openpnp-capture-java:0.0.30-1')
+    implementation('io.github.doblon8:openpnp-capture-java:0.0.1')
     implementation("io.matthewnelson.kmp-tor:runtime:2.5.0")
     implementation("io.matthewnelson.kmp-tor:resource-exec-tor-gpl:408.21.0")
     implementation('org.jetbrains.kotlinx:kotlinx-coroutines-javafx:1.10.2') {
@@ -140,6 +140,7 @@ application {
                                  "--enable-native-access=com.fazecast.jSerialComm",
                                  "--enable-native-access=org.usb4java",
                                  "--enable-native-access=io.github.doblon8.jzbar",
+                                 "--enable-native-access=io.github.doblon8.openpnp.capture",
                                  "--add-opens=javafx.graphics/com.sun.javafx.css=org.controlsfx.controls",
                                  "--add-opens=javafx.graphics/javafx.scene=org.controlsfx.controls",
                                  "--add-opens=javafx.controls/com.sun.javafx.scene.control.behavior=org.controlsfx.controls",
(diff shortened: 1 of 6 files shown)


craigraw:
Thanks for this - will get around to reviewing it soon.

craigraw:
Thanks for this PR - the FFM-based approach is a good fit now Sparrow is on JDK 25, and I appreciate that the bindings wrap the natives built from https://github.com/sparrowwallet/openpnp-capture. I've verified the bundled libraries are byte-identical to the v0.0.30-1 release assets, and that the Maven Central jar matches the tagged source.

The library code itself looks good, but the build integration has a couple of issues that mean the packaged app doesn't load the natives from the application image as intended:

1. **Missing comma in the `--exclude-resources` list.** The `glob:/io.github.doblon8.jzbar/native/**` entry has lost its trailing comma inside the concatenated string, so it fuses with the new openpnp glob into one invalid pattern - neither jzbar nor openpnp-capture natives are now excluded from the jlink image.

2. **The `nativeLibJars` extraction pattern matches nothing.** Sparrow's `osArch` is `x64`/`aarch64`, but the jar's resources are under `x86_64`/`arm64`, so `extractNativeLibraries` silently copies zero files from the jar.

3. **Nothing loads the library from the image.** The bundled natives keep their CI artifact names (e.g. `libopenpnp-capture-macos-latest-arm64.dylib`), which `System.loadLibrary("openpnp-capture")` can never resolve.

The net effect is that the webcam currently only works because 1 and 3 cancel out: the resources remain in the image, and the bindings fall back to extracting the library to a temp directory at runtime. Sparrow deliberately moved away from runtime extraction (see 98ae8918) - natives should be loaded directly from the application image.

Could I suggest the following:

- In `io.github.doblon8:openpnp-capture-java`, rename the bundled natives to standard `System.mapLibraryName` names (`libopenpnp-capture.so`, `libopenpnp-capture.dylib`, `openpnp-capture.dll`) and use the same directory convention as jzbar (`native/<os>/<x64|aarch64>/`).
- In this PR, restore the trailing comma in the exclude-resources list, and update the `nativeLibJars` entry to match the new layout.
- Also in this PR, pre-load the library from the image before first use, the same way `ZBar.loadZBar()` does. This must be the first statement in the `WebcamService` static initializer - the existing block starts with `OpenPnpCapture.setLogLevel(...)`, and the first touch of any bindings class triggers `openpnp_capture`'s own static initializer, which runs its load attempts immediately. If the pre-load comes after that, the bindings will already have fallen back to temp-dir extraction before `System.load` runs:

```java
static {
    String javaHome = System.getProperty("java.home");
    if(javaHome != null) {
        File libFile = new File(new File(javaHome, "lib"), System.mapLibraryName("openpnp-capture"));
        if(libFile.exists()) {
            System.load(libFile.getAbsolutePath());
        }
    }

    if(log.isTraceEnabled()) {
        OpenPnpCapture.setLogLevel(LogLevel.VERBOSE);
    }
    //...rest of the existing static block
}
```

(No `loaded` boolean guard is needed as in `ZBar.java` - that exists because `ZBar` loads lazily from `scan()`, which is called repeatedly, whereas a static initializer runs exactly once.)

The pre-load is necessary because `System.loadLibrary` cannot find libraries in `java.home/lib` on Windows (the JVM's `java.library.path` is captured at startup - see doblon8/jzbar#10 for the background). The bindings' static initializer already tolerates both load attempts failing and picks up the pre-loaded library via `loaderLookup()`, and the classpath fallback still covers development runs.

Two minor observations on the library, neither blocking: `CaptureResult.values()[result]` will throw `ArrayIndexOutOfBoundsException` rather than `CaptureException` if the native library ever returns an unexpected result code, and an exception escaping the FFM log upcall is fatal to the JVM where JNA used to swallow callback exceptions - probably worth a defensive try/catch in `installCustomLogFunction`.

doblon8:
Thanks for the detailed review.

I think I've addressed all the issues you pointed out, both in the library (release `0.0.3`) and on the Sparrow side.

The only point I did not change is the `nativeLibJars` entry for the new directory structure, as I believe it matches the intended structure after the changes to `openpnp-capture-java`.

Let me know if anything still looks off or if I missed something.

craigraw:
Thanks, looks good, and tested on macOS aarch64. I don't have access to all target platforms for further testing at the moment, but I don't think there will be any issues.

doblon8:
I’ve tested on Linux (Debian 13) x64 so far, and it seems to work correctly on my side as well.

I’ll also try Windows x64 soon and report back.

doblon8:
I’ve just tested it on Windows 10 x64, and everything seems to be working correctly as well.

craigraw:
Re your Linux and Windows testing - was that with the jpackaged build or `./gradlew run`?

doblon8:
It was with the jpackaged build.

I ran:
```sh
./gradlew jpackage
./sparrow -l debug -n testnet4
```
on both Linux and Windows.

craigraw:
Tested working on Linux aarch64.

craigraw:
Thanks for the work on this!

PR #2030, opened 2026-07-13
Verify scanned or loaded transactions match the originating PSBT

odudex:
When a transaction tab's Scan QR or Load Transaction buttons (or the scan button on the Show QR dialog) are used to import a signed transaction, verify it matches the transaction in the tab instead of opening a new tab on mismatch, where the discrepancy could go unnoticed before broadcasting.

### DIFF
--- a/src/main/java/com/sparrowwallet/sparrow/AppController.java
+++ b/src/main/java/com/sparrowwallet/sparrow/AppController.java
@@ -614,6 +614,10 @@ public class AppController implements Initializable {
     }
 
     public void openTransactionFromFile(ActionEvent event) {
+        openTransactionFromFile(event, null);
+    }
+
+    private void openTransactionFromFile(ActionEvent event, PSBT contextPsbt) {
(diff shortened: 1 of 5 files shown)


craigraw:
Thanks for suggesting this again. It turned out to be quite tricky, but I've implemented it in 4b8a4594 and 078af174. 

Scanned or loaded transactions will now be matched against originating tranasctions in existing tabs, and an error will displayed if no match is found. Such transactions can still be loaded from the File menu (or via drag and drop) if necessary. Loading a signed silent payments transaction from the context of an originating silent payments PSBT will now fail, as a PSBT is required for BIP375 verification of the computed output fields.

odudex:
Thank you for implementing this!
Tested here with standard and Silent Payments transactions, and it worked as expected.
IMO being strict to the specific tab from which the button was pressed would be more precise, but widening the scope to all opened tabs is OK, and still catches to more dangerous situations.

PR #2033, opened 2026-07-20
Support XDG base directory specification

### LINKED ISSUE #1897: Support XDG Base Directory Specification on Linux
Currently, Sparrow Wallet creates a `.sparrow` directory directly inside the user's home directory on Linux. This directory contains databases, configuration, logs, and other files.

Please update Sparrow Wallet to follow the XDG Base Directory Specification for storing configuration, data, and cache files. This would better integrate with Linux systems and allow users to manage their files more cleanly and in line with Linux standards.

bvotteler:
Resolves #1897 

### Usage and preconditions

Allows the use of XDG Base Directory Specification under the following circumstances:
- Applicable to Mac or Unix
- User must at the very least create the folder `XDG_CONFIG_HOME/sparrow` or if `XDG_CONFIG_HOME` is not specified, `user.home/.config/sparrow`
- System property `sparrow.home` takes precedence over XDG locations
- Default location flag takes precedence over `sparrow.home`

No existing files are migrated or changed, so the user must copy them across manually.

### File / folder locations under XDG
When using XDG specs, these are the locations hat will be used:
- `XDG_CONFIG_HOME/sparrow`: the main `config` file
- `XDG_DATA_HOME/sparrow`: wallet folders, hardware folders, certs folders, tor folders
- `XDG_STATE_HOME/sparrow`: logfile, testnet flags, lockfile
- `XDG_CACHE_HOME/sparrow`: unused

### Assumptions, feedback needed
I'd appreciate feedback for the assumptions I made to get here.
- At the very least Sparrow's XDG config folder must exist. The mere existence of environment variables alone is not sufficient.
- For any `XDG_*_HOME` env var: If it is not set, the code will use default locations under `user.home` (`.config`, `.local/share`, etc).
- The log file location was a bit tricky to decide. I went with `state` but an argument can be made to put it under `cache` instead. Particularly since it seems to grow indefinitely?
- The testnet locations are a bit of a mystery to me. I am unsure of using `state` for them and pointers would be very welcome.

And of course any other hints or feedback are much appreciated.

### DIFF
--- a/src/main/java/com/sparrowwallet/sparrow/AppController.java
+++ b/src/main/java/com/sparrowwallet/sparrow/AppController.java
@@ -525,7 +525,7 @@ public class AppController implements Initializable {
     }
 
     public void showLogFile(ActionEvent event) throws IOException {
-        File logFile = new File(SparrowDirectories.getSparrowHomeDirs().state(), "sparrow.log");
+        File logFile = new File(SparrowDirectories.getHomeDirs().state(), "sparrow.log");
         if(logFile.exists()) {
             AppServices.get().getApplication().getHostServices().showDocument(logFile.toPath().toUri().toString());
         } else {
@@ -1042,7 +1042,7 @@ public class AppController implements Initializable {
         DirectoryChooser directoryChooser = new DirectoryChooser();
         directoryChooser.setTitle("Choose Sparrow Home Folder");
         directoryChooser.setInitialDirectory(initialDir == null || !initialDir.exists()
-                ? SparrowDirectories.getSparrowHomeDirs().config()
+                ? SparrowDirectories.getHomeDirs().config()
                 : initialDir);
         File newHome = directoryChooser.showDialog(window);
 
--- a/src/main/java/com/sparrowwallet/sparrow/AppController.java
+++ b/src/main/java/com/sparrowwallet/sparrow/AppController.java
@@ -524,7 +524,7 @@ public class AppController implements Initializable {
     }
 
     public void showLogFile(ActionEvent event) throws IOException {
-        File logFile = new File(Storage.getSparrowHome(), "sparrow.log");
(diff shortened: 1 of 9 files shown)


craigraw:
Thanks very much for this @bvotteler — it's a well-considered PR, and the fact that you worked through the same questions independently was genuinely useful in confirming the approach.

I've ended up implementing XDG support directly in 2d667d11 and sparrowwallet/drongo@fa1bb4e, and have taken a different structural approach, so I'm going to close this. I want to be clear that's not a reflection on the quality of the work here.

The main reason is the log file. `logback.xml` resolves its directory through `PropertyDefiner` in drongo, which runs before any Sparrow class is loaded and has its own copy of the home-directory logic. Because drongo can't depend on Sparrow, the resolution has to live there and be called from `Storage`, rather than the other way around — so the change spans both repos. As it stands here, `sparrow.log` would continue to be written to `~/.sparrow` while Help > Show Log File looked in the XDG state directory.

A few other points, since you asked for feedback on the assumptions:

- I've made the categories resolve independently rather than gating all four on the config directory existing. That lets people move things across one at a time, which I think suits a manual migration better.
- Agreed on the log file belonging in state rather than cache. Cache should hold only what can be deleted without losing anything, and a log doesn't qualify — indefinite growth is a separate problem worth fixing on its own.
- On the network flag files: I've put them in the config directory. They're user-created files that select the network, so they're configuration, and they need to be read before the config file itself.

Thanks again for taking the time — and for the tests, which asked the right questions.

bvotteler:
No worries @craigraw , no problem at all. And thank you for taking the time to provide the feedback and pointers. They are really useful in understanding the code base better.

And interesting points about drongo. I did not notice that it loads so much earlier.

> I've made the categories resolve independently rather than gating all four on the config directory existing. That lets people move things across one at a time, which I think suits a manual migration better.

Love this! That's extremely flexible.

PR #2035, opened 2026-07-28
Add option to ignore dust transactions on private key sweep

nroktib:
### Summary
Add an option to the sweep private key dialog to set a minimum sat threshold for UTXOs.

Screenshot from 2026-07-27 22-05-47


craigraw:
This is quite a bit of additional UI complexity. Would the configured dust threshold (and just the toggle in the UI) not be sufficient?

nroktib:
Sure, like this?
I was looking at the SP value, is it worth adjusting for that? Then the question is, adjust based on the sweep address or from pk? Can you even tell if the from pk is an SP?

nroktib:
Sorry, made a mess of this commit, moved to #2036

PR #2036, opened 2026-07-28
Ignore dust on sweep

nroktib:
### Summary

Add an option to the sweep private key dialog to ignore dust UTXOs.

Screenshot from 2026-07-28 08-39-55

### DIFF
--- a/src/main/java/com/sparrowwallet/sparrow/control/PrivateKeySweepDialog.java
+++ b/src/main/java/com/sparrowwallet/sparrow/control/PrivateKeySweepDialog.java
@@ -27,6 +27,7 @@ import com.sparrowwallet.sparrow.io.CardApi;
 import com.sparrowwallet.sparrow.io.Config;
 import com.sparrowwallet.sparrow.net.ElectrumServer;
 import com.sparrowwallet.sparrow.net.ServerType;
+import com.sparrowwallet.sparrow.control.UnlabeledToggleSwitch;
 import javafx.application.Platform;
 import javafx.collections.FXCollections;
 import javafx.event.ActionEvent;
@@ -68,6 +69,7 @@ public class PrivateKeySweepDialog extends Dialog<Transaction> {
     private final ComboBox<Wallet> toWallet;
     private final FeeRangeSlider feeRange;
     private final CopyableLabel feeRate;
+    private UnlabeledToggleSwitch useDustLimit;
     private SilentPaymentAddress silentPaymentAddress;
 
     public PrivateKeySweepDialog(Wallet wallet) {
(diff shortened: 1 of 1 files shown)


craigraw:
The code needs a cleanup: Remove the unnecessary import, all comments, additional blank lines, and all `log.info` lines (no logging is required). Also check the code formatting - some places lack spaces (`x=y`).

craigraw (commented):
Thanks for this — sweeping a dusted key without pulling in the attacker's UTXOs is a useful addition. A few things need addressing before this can go in.

**Dust filtering must happen before the silent payment address is computed**

In `createTransaction(ECKey, ScriptType, List<TransactionOutput>, Payment)`, `destAddress` is computed on the first line, but the filter is applied after it:

```java
Address destAddress = payment instanceof SilentPayment silentPayment ? computeSilentPaymentAddress(privKey, scriptType, txOutputs, silentPayment) : payment.getAddress();
ECKey pubKey = ECKey.fromPublicOnly(privKey);

if(useDustLimit.isSelected()) {     // too late
    txOutputs = removeDust(txOutputs);
}
```

`computeSilentPaymentAddress()` derives the address from the outpoints of every UTXO passed to it (BIP-352 `input_hash`). Removing inputs afterwards means the broadcast transaction's inputs no longer match the set the address was derived from, so the recipient scanning for the payment computes a different tweak and never detects the output. Nothing catches this — the output is added as a plain script without PSBT silent payment metadata, so `PSBT.extractTransaction()` skips `validateSilentPayments()`. The funds would be unrecoverable.

**Handle the case where everything is filtered out**

If all the UTXOs are dust, the list ends up empty and the user gets "The unspent outputs for this private key contain insufficient funds to spend (0 sats)", which is misleading. It also means `computeSilentPaymentAddress()` would be handed an empty outpoint set.

Both issues are best fixed in the caller, so the whole method works off one consistent list and the parameter isn't reassigned:

```java
ElectrumServer.AddressUtxosService addressUtxosService = new ElectrumServer.AddressUtxosService(fromAddress, since);
addressUtxosService.setOnSucceeded(successEvent -> {
    List<TransactionOutput> utxos = addressUtxosService.getValue();
    if(ignoreDust.isSelected()) {
        utxos = removeDust(utxos);
        if(utxos.isEmpty()) {
            AppServices.showErrorDialog("No outputs to sweep", "All of the unspent outputs for this private key have been ignored as dust.");
            return;
        }
    }

    createTransaction(privateKey.getKey(), scriptType, utxos, payment);
});
```

**Smaller items**

- `import com.sparrowwallet.sparrow.control.UnlabeledToggleSwitch;` is the class's own package — remove it.
- `useDustLimit` should be `private final` and declared with the other controls; every other control field in the class is final.
- Name it `ignoreDust` to match the label and the behaviour.
- `setSelected(false)` is redundant, `ToggleSwitch` is unselected by default.
- `removeDust()` reads better as a stream, consistent with the rest of the class:
  ```java
  private List<TransactionOutput> removeDust(List<TransactionOutput> txOutputs) {
      long dustAttackThreshold = Config.get().getDustAttackThreshold();
      return txOutputs.stream().filter(txOutput -> txOutput.getValue() > dustAttackThreshold).collect(Collectors.toList());
  }
  ```
- Note the boundary there: `WalletUtxosEntry.calculateDust()` flags UTXOs with `<= dustAttackThreshold`, so use `>` here to keep the same UTXO set that's marked as a dust attack in the UTXO table.

nroktib:
Appreciate the education on SP and the updates.

craigraw:
Thanks!

PR #2038, opened 2026-07-30
Add list of toxic utxos to remove from sweep txns

nroktib:
Allows users to add a list of txids to exclude on sweeps by adding the following to their config file.

"toxicTransactions": [
    "txid1....",
    "txid2....",
    "txid3...."
  ]

### DIFF
--- a/src/main/java/com/sparrowwallet/sparrow/control/PrivateKeySweepDialog.java
+++ b/src/main/java/com/sparrowwallet/sparrow/control/PrivateKeySweepDialog.java
@@ -69,6 +69,7 @@ public class PrivateKeySweepDialog extends Dialog<Transaction> {
     private final FeeRangeSlider feeRange;
     private final CopyableLabel feeRate;
     private final UnlabeledToggleSwitch ignoreDust;
+    private final UnlabeledToggleSwitch includeToxicTxns;
     private SilentPaymentAddress silentPaymentAddress;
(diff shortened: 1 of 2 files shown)


craigraw:
This feels very niche, and has an awkward UX. What is the motivation here?

nroktib:
Agreed, it's is, and it's a little clunky, not for everyone...

As it stands there is no easy way to drop utxos during a sweep, this is an "easy" way to lock them out of all sweeps, dust is a little subjective, this is absolute.

The main motivation is for things like this; [Galaxy Insights](https://www.galaxy.com/insights/research/bitcoin-wallet-opreturn-salomon-brothers-abandoned-property-constructive-possession)

There are 8 toxic transactions linked in that article, there are others like it... These 8 could be set as the defaults and updated, or just left as is and up to the user.

A date might be simpler, ie. the linked stuff all happened in July 2025, so dropping anything after July 1 2025 would keep them out but I couldn't find a way to easily link the utxo to it's date, also prone to issues if the wallet was used after the dusting.

Maybe we could lock it behind an "expert mode" (just committed) that way the average user doesn't see it?

craigraw:
I am not a fan of "expert modes" in general. 

I appreciate the suggestion, but I think adding code for a niche use case in legacy single address wallet technology is probably not on balance a good idea.

PR #2042, opened 2026-08-04
Fix Electrum read timeout not actually bounding the wait for a response

noop32:
## Bug

`TcpTransport.readResponse()` only bounded **acquiring `readLock`**
via `tryLock(timeout)`. `readLock` is free whenever the reader thread
is blocked on the socket (it releases the lock while
awaiting/reading), so `tryLock` returns almost immediately regardless
of the requested timeout — the actual wait for a response is the
**unbounded** `readingCondition.await()` right below it.

`readLine()` also swallows `SocketTimeoutException` in a retry loop
and never gives up on its own. So a server that accepts a request and
never replies hangs the caller forever, with every other RPC queued
behind `clientRequestLock`. The transport was also never marked
failed on the (effectively unreachable) timeout path, so Sparrow never
reconnected.

Knock-on effect: `readTimeoutIndex` — and the page-halving backoff in
`PagedBatchRequestBuilder.getPageSize()` built on top of it via
`TimeoutCounter` — essentially never escalates in this scenario, since
the lock timeout meant to drive it is effectively unreachable.

## Fix

- Track a single time budget across both the `tryLock()` and the
  subsequent wait, using the idiomatic
  `Condition.awaitNanos()` loop instead of an unbounded `await()`.
- On genuine expiry, actually fail the connection (stop the reader
  loop, close the socket) instead of only logging/escalating, so the
  failure surfaces to callers and existing reconnect handling kicks in
  instead of the connection sitting open indefinitely.

Pre-existing decay quirk (`readTimeoutIndex--` when already at max,
right after a successful lock acquisition) is left untouched — out of
scope for this fix.

No build toolchain available in the environment this was written in,
so this has been reviewed by hand/tracing rather than compiled against
the full project — happy to iterate on review feedback.

### DIFF
--- a/src/main/java/com/sparrowwallet/sparrow/net/TcpTransport.java
+++ b/src/main/java/com/sparrowwallet/sparrow/net/TcpTransport.java
@@ -144,29 +144,48 @@ public class TcpTransport implements CloseableTransport, TimeoutCounter {
             }
         }
 
+        //This budget must bound the ENTIRE wait for a response - not just acquiring
+        //readLock below, but also the subsequent wait for the reader thread to deliver
+        //data. Previously only lock acquisition was time-bounded here; the
+        //readingCondition.await() further down was unbounded. readLock is free whenever
+        //the reader thread is blocked on the socket (it releases the lock while
+        //awaiting/reading), so tryLock below returned almost immediately regardless of
+        //the requested timeout, and a server that accepted a request and never replied
+        //hung the caller on that unbounded await forever - never timing out, never
+        //failing the connection, and never triggering a reconnect. Every subsequent RPC
+        //then queued forever behind clientRequestLock too.
+        long remainingNanos = TimeUnit.SECONDS.toNanos(readTimeouts[readTimeoutIndex])
+                + TimeUnit.MILLISECONDS.toNanos(requestIdCount * PER_REQUEST_READ_TIMEOUT_MILLIS);
+
+        boolean locked;
         try {
-            if(!readLock.tryLock((readTimeouts[readTimeoutIndex] * 1000L) + (requestIdCount * PER_REQUEST_READ_TIMEOUT_MILLIS), TimeUnit.MILLISECONDS)) {
-                readTimeoutIndex = Math.min(readTimeoutIndex + 1, readTimeouts.length - 1);
-                log.warn("No response from server, setting read timeout to " + readTimeouts[readTimeoutIndex] + " secs");
-                throw new IOException("No response from server");
-            }
+            locked = readLock.tryLock(remainingNanos, TimeUnit.NANOSECONDS);
         } catch(InterruptedException e) {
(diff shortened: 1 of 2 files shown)


noop32:
Will investigate further

noop32:
Added a regression test (`TcpTransportTimeoutTest`) and verified E2E:

- **pre-fix** `TcpTransport`: test hits `AssertTimeoutPreemptively`'s timeout — confirms the hang described above actually reproduces.
- **post-fix**: passes in ~3.7s, timeout escalates 3s→8s (log: `No response from server, setting read timeout to 8 secs`), and the connection is correctly marked failed afterward.

Built/tested with JDK 25.0.4 + the Gradle 9.1.0 wrapper: `./gradlew :test --tests com.sparrowwallet.sparrow.net.TcpTransportTimeoutTest`.

noop32:
Rebased onto current upstream `master` and re-ran the full E2E check to make sure this wasn't validated against a stale checkout:

- **tested against**: `sparrowwallet/sparrow@b0b9cc23d4e4a38107cf50469723fc7f550ce00a` ("reject truncated and oversized tlv lengths when parsing keycard responses", 2026-08-04 14:31 UTC)
- `net/` package untouched by the 10 commits between the original fork point and this HEAD, so the fix itself needed no changes — only `drongo`/`lark` submodule pins moved
- `./gradlew :test --tests com.sparrowwallet.sparrow.net.TcpTransportTimeoutTest` → `BUILD SUCCESSFUL`, 1 test, 0 failures, 0 errors

noop32:
Split the BitcoindTransport fix out into its own PR: #2043 (different transport, independent fix — this PR now covers only the Electrum TcpTransport timeout fix).

craigraw:
Thanks — the diagnosis is right, and it's a regression I introduced in 273d2aac. Before that commit the read thread held `readLock` while blocked on the socket, so `tryLock` genuinely bounded the response wait. Once it released the lock during reads, the only real wait became the unbounded `await()`. Your test fails on master and passes with the patch, so the read timeout has been dead since May. Good catch.

I'm taking the first half. I'm not taking the socket close, though - a read timeout here is deliberately non-fatal. `RetryLogic` retries over the same transport, `TimeoutCounter` halves the batch page size, and the stale-id discard loop in `pass()` exists to absorb the late response. Failing the connection means a server that replies 2s past the timeout gets torn down and the remaining retries fail against a dead socket. The "connection sits open indefinitely" case that motivated it is a symptom of the unbounded await: once the wait is bounded, a server that never replies escalates 3→8→16→34 across the retries and surfaces `ServerException` → reconnect on its own.

So `failReadTimeout()` becomes just the escalate-and-log it replaced. I also dropped the `isConnected()` assertion, shortened the timeouts the test waits out, and added a companion test that a late-replying server recovers on the next request.

Implemented in 5b10debd.

PR #2043, opened 2026-08-04
Fix BitcoindTransport hanging forever when Bitcoin Core RPC never replies

noop32:
HttpURLConnection defaults both connect and read timeout to 0 (infinite), and neither was ever set here, so an unreachable/black-holed bitcoind (getOutputStream, which triggers the connect) or one that accepts the connection but never responds (getResponseCode) hangs the caller — and whatever thread called `pass()` — forever.

- Connect timeout: 15s direct, 60s when routed over a Tor proxy to an onion address (onion circuit builds are slow — mirrors the precedent `TcpTransport` already sets via `SLOW_READ_TIMEOUT_SECS` for `BITCOIN_CORE` + onion).
- Read timeout: 60s for all methods except `importdescriptors`, which is exempted because a wallet-birthday rescan can legitimately block for hours. Method name is parsed out of the request JSON (not string-matched), so a `params` value cannot spoof the allowlist.

Tested with a bare, non-responding HTTP server standing in for a black-holed bitcoind: pre-fix, `pass()` hangs indefinitely (confirmed via a differential probe against the pre-fix class); post-fix, it fails within the read timeout instead. Also verified `importdescriptors` is correctly exempted and that the method-name JSON parsing is not fooled by an unrelated `params` value containing the string.

Same failure class as #1 / sparrowwallet#2042 (untimed `HttpURLConnection`/read loop letting a non-responding server hang the client forever), split out into its own PR since it is a different transport (Bitcoin Core RPC vs Electrum TCP) with an independent fix.

### DIFF
--- a/src/main/java/com/sparrowwallet/sparrow/net/cormorant/bitcoind/BitcoindTransport.java
+++ b/src/main/java/com/sparrowwallet/sparrow/net/cormorant/bitcoind/BitcoindTransport.java
@@ -1,5 +1,8 @@
 package com.sparrowwallet.sparrow.net.cormorant.bitcoind;
 
+import com.fasterxml.jackson.core.JsonFactory;
+import com.fasterxml.jackson.core.JsonParser;
+import com.fasterxml.jackson.core.JsonToken;
 import com.github.arteam.simplejsonrpc.client.Transport;
 import com.sparrowwallet.drongo.Network;
 import com.sparrowwallet.sparrow.AppServices;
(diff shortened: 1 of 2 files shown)


craigraw:
Thanks - the diagnosis is right. Going in with some changes:

- Read timeout raised to 300s, floored at the configured max server timeout. These calls serve Electrum requests that `TcpTransport` already bounds at up to 208s for an onion node, so bounding the inner leg more tightly only fails the request sooner. It is a backstop against a hung thread, not a limit on how long the user waits.
- `loadwallet` exempted alongside `importdescriptors`, and the set renamed `RESCANNING_METHODS`. Core rescans synchronously in `AttachChain` from the block the wallet was last processed at to the tip. Timing it out is worse than the hang: `initialize()` catches only `JsonRpcException`, so the retry hits a wallet still loading and throws "Wallet already loading" uncaught from `getListSinceBlock`, failing every reconnect for the length of the rescan.
- Connect timeout now taken from the proxy `pass()` already resolved, rather than calling `AppServices.getProxy()` a second time and setting another default `Authenticator`.

Implemented in e49b6a4a.

Issue #622, opened 2022-07-13
Derive BIP85 child seeds


orangesurf:
(Originally posted on telegram chat on 7th July)

With BIP85 both the index and the wallet type are required for recovery. There are at least 8 options with the current coldcard firmware
- BIP39 12 Words
- BIP39 18 Words
- BIP39 24 Words
- WIF (private key)
- XPRV (BIP-32)
- hex
- 32 bytes
- 64 bytes

A wallet import option to scan the possible child seeds for a known index number would help greatly when the wallet types have been lost.

Where a number is not known it is may be typical that indexes were used sequentially starting from 0. Here an option to scan index numbers sequentially (starting from 0 until a gap limit of empty child wallet indexes is reached) would be advantageous.

Transisto:
https://github.com/bitcoin/bips/blob/master/bip-0085.mediawiki

PR #2050, opened 2026-08-10
Add an option to derive a BIP85 child mnemonic when importing BIP39 software wallet

elmeriniemela:
Add an option to derive a BIP85 child mnemonic when importing BIP39 software wallet. The scope is intentionally narrowed to only BIP39 software keys to make the diff as small as possible. Future PR's can add derivation for other  key types. 

A new button to derive "BIP85 child mnemonic"

1pw

Dialog asks for the word count and index. Optional parent passphrase needs to match. With the passphrase, an additional warning to verify the master key fingerprint is shown to the user.

2pw

After deriving the BIP85 child, the process continues normally by asking the user to "Confirm backup" of the child mnemonic.

3pw

The example from the screenshots was crosschecked against a SeedSigner. Further correctness tests are part of the drongo PR linked below.

Related links
- drongo: https://github.com/sparrowwallet/drongo/pull/47
- earlier issue: https://github.com/sparrowwallet/sparrow/issues/622

Disclaimer: LLM's were used as part of the development process. The code changes were iterated, reviewed and manually tested by me.

### DIFF
--- a/src/main/java/com/sparrowwallet/sparrow/control/MnemonicKeystoreImportPane.java
+++ b/src/main/java/com/sparrowwallet/sparrow/control/MnemonicKeystoreImportPane.java
@@ -1,25 +1,37 @@
 package com.sparrowwallet.sparrow.control;
 
 import com.sparrowwallet.drongo.KeyDerivation;
+import com.sparrowwallet.drongo.Utils;
 import com.sparrowwallet.drongo.crypto.ChildNumber;
+import com.sparrowwallet.drongo.crypto.DeterministicKey;
+import com.sparrowwallet.drongo.crypto.HDKeyDerivation;
+import com.sparrowwallet.drongo.crypto.HDDerivationException;
 import com.sparrowwallet.drongo.protocol.ScriptType;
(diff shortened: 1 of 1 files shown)


craigraw:
Unfortunately I feel this proposal adds too much UI complexity to the common use case of creating a software wallet in Sparrow. 

As a separate minor point I don't think creating BIP85 seeds in Sparrow is a compelling use case, as opposed to hardware wallets where it has more value. But the first point is the material one.

elmeriniemela:
Thanks for the reply! I had a feeling that you might oppose adding this to the UI. :)

I would love to have BIP85 derivation available somewhere in Sparrow, even if it were hidden behind an “Advanced” option. In my use case, all the other signers in my multisig setup allow me to derive BIP85 child seeds, except Sparrow.

So, to create a similar multisig setup for someone else using BIP85-derived seeds, I have to move my Sparrow seed to different software. This means exposing the parent seed either to a new piece of software or to the same software that I used to create one of the other seeds in my multisig setup. I feel that both options slightly weaken the overall security and separation of my setup.
 
 Is there anywhere I could move the feature to make it acceptable?

Issue #1672, opened 2025-03-25
https can't be used with a bitcoin core server when not on port 443


w3irdrobot:
when setting the server host, it appears the protocol [gets chopped off](https://github.com/sparrowwallet/sparrow/blob/7fa13901d4f2aaade5ca125e57eb0bdabec64fa1/src/main/java/com/sparrowwallet/sparrow/settings/ServerSettingsController.java#L779-L780). Then when the server is saved in the config, it's [saved with the `HTTP` protocol](https://github.com/sparrowwallet/sparrow/blob/7fa13901d4f2aaade5ca125e57eb0bdabec64fa1/src/main/java/com/sparrowwallet/sparrow/settings/ServerSettingsController.java#L801-L802) unless the port 443 is specified. my expectation would be to be able to specify a url that includes the protocol if i don't want the default of `HTTP` and then specify a non-standard port to connect to. 

i'm willing to do the work. i wanted to make sure it's desired before starting on it.

related: #774

craigraw:
As you can see from the linked issue, I don't want to show a full protocol here to keep the UI simple and consistent. One way to address this would be to allow pasting in of a URL like `https://xyz.abc:8443`. Currently, this will set a `HTTP` URL in the config, but it could be made to set `HTTPS` with a little work.

Also worth noting: while I'd like to support this edge case, I'm in general not a fan of exposing Bitcoin Core RPC over an untrusted network.

w3irdrobot:
I think omitting the protocol might create some confusion when HTTP isn't the assumed default. HTTPS has become quite standard for most internet interactions these days, even if it's not yet the norm in bitcoin applications. If there are UI concerns about displaying the protocol, that's understandable, and I'm glad you're generally agreeable to adding HTTPS support given its widespread use elsewhere. I'd be willing to help implement this feature if that would be helpful.

> Also worth noting: while I'd like to support this edge case, I'm in general not a fan of exposing Bitcoin Core RPC over an untrusted network.

Would you mind sharing your thoughts on this approach? While I see the value in encouraging users to run their own nodes, there's the practical consideration of making those nodes accessible from different locations. In cases where someone needs to access their bitcoin node remotely (like when hosted on a public VPS), HTTPS would offer some additional protection for data in transit.

craigraw:
> Would you mind sharing your thoughts on this approach?

[...]

Issue #774, opened 2022-12-13
[Feature Request] Use an URL instead of hostname to connect to a RPC server


NicolasDorier:
### Issue

As you can see here, while the field say `URL`, an URL doesn't work.

![image](https://user-images.githubusercontent.com/3020646/207211884-48e5b6d0-bbf0-4571-915e-8341dd4a1d70.png)

### Why should an URL be supported

I think an URL should be supported. I am planning to create some middleware to allow users of BTCPay an easy way to access their wallet through Bitcoin Core RPC. But if I do so, the URL will look like this:

`https://mainnet.demo.btcpayserver.org/rpc` 

By doing so we can use reverse proxy to route the request to the middleware (or even directly to core), without having to ask user to open new port.

This mean that rather than two field, hostname and port, I wished to only have one field URL.

craigraw:
> As you can see here, while the field say URL, an URL doesn't work.

Yes, `Host & Port` would be more accurate, but also arguably more technical language.

> This mean that rather than two field, hostname and port, I wished to only have one field URL.

That's a fairly significant change to the UI, and one I would argue adds complexity for the default use case. The current "host & port" field format is used for configuring Electrum and proxy servers as, so the consistency helps. It may be possible to support this via pasting in a http:// URL, although the `/rpc` path is not easily representable even if it is stored in the configuration.

> I am planning to create some middleware to allow users of BTCPay an easy way to access their wallet through Bitcoin Core RPC.

Do you mean allowing access to Sparrow to create it's own Bitcoin Core wallet in the instance hosted by BTCPay?

NicolasDorier:
@craigraw when a merchant create a wallet on BTCPay, he can choose to add the generated descriptors into Bitcoin Core.
We then expose Bitcoin core RPC over Tor v3, so people can use it with third party wallet such as Sparrow.

The problem is that it only works on Tor, I wished we could make it works on HTTPS as well so it is easier and more reliable to use.

The only alternative, if you don't support HTTP path would be for user to setup a separate domain for Bitcoin RPC, which worse UX than what we have with Tor currently.

craigraw:
> We then expose Bitcoin core RPC over Tor v3

I'm actually not a fan of exposing Bitcoin Core RPC over any untrusted network - if one discovers the host name all wallet privacy is at risk, and an attacker can take control over the node itself. The authentication provided is not particularly well defended. That said I'm willing to support it so long as it doesn't make the default, recommended approach a poorer UX.

[...]

PR #2068, opened 2026-08-27
retain the entered scheme when configuring a bitcoin core server

MattDHill:
The Bitcoin Core server's scheme is derived from the port alone, so a pasted `https://node.example:8332` is stored as `http://node.example:8332`, and the TLS transport and certificate pinning added in 2.4.2 are reachable only on 443 — which is not usually where a private node with a self-signed certificate sits.

This retains a scheme the user entered: from a pasted URL, from the configured server when the view opens, and from a recent or QR scanned server. Port 443 still implies HTTPS, so no existing configuration changes behaviour, and there is no new UI, per the discussion in #1672 and #774.

The scheme is sticky once entered, since neither field displays it. Pasting a URL carrying the other scheme switches it back.

### DIFF
--- a/src/main/java/com/sparrowwallet/sparrow/settings/ServerSettingsController.java
+++ b/src/main/java/com/sparrowwallet/sparrow/settings/ServerSettingsController.java
@@ -177,6 +177,8 @@ public class ServerSettingsController extends SettingsDetailController {
 
     private boolean coreServerWarningShown;
 
+    private Protocol coreProtocol;
+
     @Override
     public void initializeView(Config config) {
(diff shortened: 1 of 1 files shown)


craigraw:
Thanks for the PR - I've gone a slightly different route with a protocol choice saved on a host key. Also see https://github.com/sparrowwallet/sparrow/issues/1672#issuecomment-5451962974.

MattDHill:
Keying on the host is better than what I did — mine kept the scheme across a host change. Thanks for reconsidering, and for the NPE fix; that was the error our users were actually hitting.

PR #2077, opened 2026-09-15
Support importing Bitcoin Core wallet.dat files

MrHodlX:
EntropyLab currently exports a ready-to-load Bitcoin Core wallet.dat for single-signature watch-only wallets, and a multisig version is about to merge. It would be great if Sparrow could accept that same file directly, so users don't have to manually copy descriptors. Sparrow already parses output descriptors, so the work is mostly reading the Core SQLite database format and feeding the descriptors inside into the existing importer. This covers both single-sig and the upcoming multisig watch-only files without touching private keys.


craigraw:
Thanks, but reading Bitcoin Core's `wallet.dat` is out of scope for Sparrow. It's Core's internal storage database rather than a wallet interchange format, and supporting it would mean adding SQLite (and Berkeley DB for older files) to parse records that can change between Core releases.

Descriptors are already the interchange format here. Sparrow imports a plain text file containing an output descriptor (File → Import Wallet → Output Descriptor), including multisig and multipath `<0;1>` descriptors, and can also scan one from a QR code. If EntropyLab exports the descriptor as a text file alongside the `wallet.dat`, it can be imported into Sparrow directly, and into most other wallets too.

Closing this, as it's a proposal rather than an implementation.

